PR #21, opened 2022-04-11
Allow full Behat runs with boost_union fixes #14.

### LINKED ISSUE #14: Feature: allow Behat full runs
If core renderers and functions are overriden and therefore navigation in theme Boost Union works differently than in Boost, then Behat handlers have to be adopted, too, to allow.
php admin/tool/behat/cli/init.php -a boost_union and then a successful run of all the present Moodle suite Behat tests.

lucaboesch:
Allow a full Moodle Behat run with Boost Union.

That adresses issue #14.

To have the Behat tests execute with the Boost Union theme you build the scenarios for the suite with
`php admin/tool/behat/cli/init.php -a boost_union`
and then run
`php admin/tool/behat/cli/run.php --tags=@core --suite="boost_union"`.

(That is just an example, omit --tags=@core for a full run.)

Best,
Luca


abias:
Hi @lucaboesch ,

thank you for working on this improvement.

I understand this PR as the basis for future implementations where theme_boost_union will deviate from theme_boost. But currently it is perfectly ok that the added test class files are practically empty.

Relying on your experience in this field, I would now just merge this in, but I have two minor things which I would like to ask you first:

1. Could you please just increment the plugin version in version.php by the last digit, i.e. the new version should be 2022031701. This ist just to be prepared for the future when this theme might be supported for multiple Moodle major versions at the same time.

2. Wouldn't it make sense to run the Github Action Workflow with this theme as well? Shouldn't we add the suite to the 
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/.github/workflows/moodle-plugin-ci.yml#L107 line?

Cheers,
Alex

lucaboesch:
Those are very good points, @abias.
I addressed those two.
Best,
Luca

abias:
Thank you very much, @lucaboesch !

I just merged your PR into the master branch. I just had to merge it manually as I had to raise the version number once more as another patch with a version incrementation had been merged into master before.

Issue #2, opened 2022-03-17
Prepare settings.php page


abias:
The settings.php page currently contains a copy of the Boost settings as the absolute minimum code needed for a Boost Child plugin. 

The next step would be to to structure the settings page in a way that more features and settings can be easily integrated into Boost Union.

PR #26, opened 2022-06-20
Prepare settings.php page, solves #2

abias:
This patch prepares the settings.php page for the integration of all the upcoming Boost Union settings.
To achieve this, it slightly breaks up the existing settings code from Boost Core while keeping all existing settings.
As a side effect, it adds lazy loading to the existing setting strings.


abias:
As this patch is more or less a no-brainer and a blocker to other upcoming patches, I dare to merge this PR without official peer review.

Issue #7, opened 2022-03-17
Feature: Back to top button


abias:
Moodle 4.x got rid of the "Back to top button". The fact why this was done is briefly noted in https://docs.moodle.org/dev/Moodle_4.0_developer_update#Removal_of_back_to_top_link.

However, some institutions might still need want to have a "Back to top" button for some reason.

It would be quite easy to add an admin setting to Boost Union to provide a "Back to top" button again.

tjettmers:
It is currently discussed as a possible solution for https://tracker.moodle.org/browse/MDL-74223, which actually belongs to https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/1

see https://www.figma.com/proto/0NPKDxZF1zEX7I0WQd1lyL/4.1-UX-LMS?page-id=2453%3A79730&node-id=4906%3A171779&viewport=16471%2C3465%2C0.65&scaling=min-zoom&starting-point-node-id=2522%3A72401

abias:
I had a look at the mockup in MDL-74223 and have implemented what can be seen there in PR #31 .

If Moodle core ever add this back-to-top button back, the functionality can be removed again from Boost Union.

PR #31, opened 2022-07-01
Feature: Back to top button, solves #7

abias:
This patch adds an admin setting which allows the admin to add a back to top button to the page. The button will appear in the bottom right corner of the page as soon as the user scrolls down the page.

A button like this existed already on Boost in Moodle Core until Moodle 3.11, but was removed in 4.0. With Boost Union, you can bring it back.


Thendo-T (commented):
I know very little of php, but I'm loving this update 👍🏾

lucaboesch (approved):
I've tested the pull request.
Looking good!

Issue #28, opened 2022-06-23
Allow non-admins to edit theme settings


abias:
While we add more and more settings to the theme and as many settings are settings which should be set rather by a manager than an admin, it would be worthwhile to allow non-admins to edit theme settings as well.

Unfortunately, Moodle core does not provide a clear path for this goal at the moment and the discussion on https://tracker.moodle.org/browse/MDL-50377 did not trigger a real result.

This issue is to examine possible workarounds for Boost Union.

abias:
In the end, it was easier than expected (and the discussion in MDL-50377 turned out not to be really constructive). I built a patch for this feature and contributed it in PR #40.

abias:
Closing as #40 is merged

PR #40, opened 2022-07-01
Feature: Allow non-admins to edit theme settings, solves #28

abias:
This patch introduces the theme/boost_union:configure capability which allows managers to configure the theme as non-admin. By default, the capability is assigned to no role at all.

----

Please note that this PR is built on top of PR #31 and should be merged in line.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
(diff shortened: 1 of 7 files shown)


lucaboesch (approved):
I've reviewed and checked this pull request.
It is looking all good.
Merging.

Issue #46, opened 2022-07-07
Feature: Show a warning banner if JavaScript is disabled


abias:
Moodle has been practicing progressive enhancement for years. However, it does not do this anymore. Even worse, it silently relies on JavaScript being enabled in the browser and does not output any notice if it is not. But if JavaScript _is_ disabled for any reasons, the user will have widgets which simply don't work.

I propose to add a warning banner which clearly tells the user to enable JavaScript if it is off.

lucaboesch:
To be honest I find this a bit absurd.
Nowadays you pratically need a computer scientist's degree to disable javascript in a default browser, don't you think?
And then, i can assume, you should be able to forsee the consequences, shoudn't you?
But if you really want this, all right then.

abias:
I agree that the standard user will never even think about disabling JavaScript in the browser. And you are right, it has become more difficult to disable JS as the browser vendors have moved the necessary switched to the back of the browser machine room.

However, please do not forget the advanced users who use browser extensions like NoScript (https://addons.mozilla.org/de/firefox/addon/noscript/) and who want to selectively enable JS on a per-site basis.

In the end, this was a request by one of our customers who told us that they _have_ such users and the actively want to inform them that JS is disabled.

lucaboesch:
Done

PR #47, opened 2022-07-07
Feature: Show a warning banner if JavaScript is disabled, solves #46

abias:
With this patch, the admin can enable a hint which will appear at the top of the Moodle page if JavaScript is not enabled. This is particularly helpful as several Moodle features do not work without JavaScript.

----

Please note that this PR is built on top of PR #45 and should be merged in line.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2022-07-07 - Feature: Show a warning banner if JavaScript is disabled, solves #46
 * 2022-07-07 - Feature: Make the course content column width configurable, helps to resolve #18.
 * 2022-07-06 - Feature: Built-in imprint, solves #32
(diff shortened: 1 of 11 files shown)


lucaboesch (changes requested):
Dear @abias
I'm afraid I don't agree with the text-center class for that warning banner.
Although I know that it will probably be shown only on very very rare occasions (people having turned off JavaScript) I think there should be consistency with the other banners.

lucaboesch:
Ordinary alignment works well, too.

If certain client want it to be styled this way though, maybe you could think of giving it an id and then styling it with custom SCSS.

grafik

abias:
That's fine, Luca. I removed the alignment class and force-pushed the branch.

lucaboesch:
Thanks, @abias !
Merged.

Issue #44, opened 2022-07-06
Release: Add contributors to README.md


abias:
Before the release, it should be discussed if we should add all institutions who have funded contributions to this plugin to README.md

lucaboesch:
BFH University of Applied Sciences, Luca Bösch: Code, Peer Review, Ideating
lern.link, Alexander Bias: Code, Peer Review, Ideating, Funding
University of Lübeck, Christian Wolters: Peer Review, Ideating
ZHAW Zurich University of Applied Sciences: Funding

lucaboesch:
Maybe in a form (although it's kind of obvious), Moodle an Hochschulen e.V. should be mentioned, too.
Any idea on how to do this?
Or leave it away, since it's self-explaining?

lucaboesch:
Additionnally, I would vote to leave the "Credits" section in

Credits

This theme is a successor of and heavily inspired by the former theme theme_boost_campus by Kathrin Osswald and Alexander Bias from Ulm University which was maintained until Moodle 3.11 on https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus.

and only then add

Contributors

…
…
÷

abias:
Hi Luca,

thanks for your numerous thoughts!

> I prefer "Contributors" over "Credits"

Yes, I agree with that and have changed the title of this issue accordingly.

> And I could imagine a specification of the work/contributions.

I agree with that as well!

> Maybe in a form (although it's kind of obvious), Moodle an Hochschulen e.V. should be mentioned, too.

I do not think that this is necessary as Moodle an Hochschulen e.V. is the publisher of this theme.

> Additionnally, I would vote to leave the "Credits" section in

Sure!

abias:
As we approach a first release, I just went ahead and have added a 'contributors' section based on the discussion above to README.md in ccf327f.

PR #48, opened 2022-07-08
Move back to top button when right drawer is shown

yanus:
Currently the "Back to top" button floats over the right drawer when it is shown (see screenshot).
This pull request moves it over the main content to line up with the floating footer button.

image

### DIFF
--- a/scss/boost_union/post.scss
+++ b/scss/boost_union/post.scss
@@ -21,10 +21,16 @@
     }
     /* Make sure that the back to top button is not covered by anything else. */
     z-index: $zindex-dropdown;
+    /* Animate position together with footer button. */
+    @include transition(0.2s);
 }
 #back-to-top i::before {
     /* Move the caret icon slightly up for a nicer look. */
     vertical-align: 0.3rem;
+}
+#page.drawers.show-drawer-right #back-to-top {
+    /* Move the back to top button when right drawer is shown. */
+    right: calc(#{$drawer-right-width} + 2rem);
 }
 
 


lucaboesch:
@abias maybe you could peer review this?

scss/boost_union/post.scss
```diff
@@ -21,11 +21,17 @@
...
+    /* Animate position together with footer button. */
+    transition: right 0.2s;
 }
 #back-to-top i::before {
     /* Move the caret icon slightly up for a nicer look. */
     vertical-align: 0.3rem;
 }
+#page.drawers.show-drawer-right #back-to-top {
+    /* Move the back to top button when right drawer is shown. */
+    right: calc(315px + 2rem);
```
abias, line 33: Looking at the place where Moodle core sets this style for the footer button on https://github.com/moodle/moodle/blob/master/theme/boost/scss/moodle/layout.scss#L212-L214, you will see that these 315px come from a SCSS variable.

So, this line should say:

```suggestion
    right: calc(#{$drawer-right-width} + 2rem);
```

PS: It's not necessary to add the .jsenabled class as well as the back-to-top button is not shown at all without JS enabled.

scss/boost_union/post.scss
```diff
@@ -21,11 +21,17 @@
     }
     /* Make sure that the back to top button is not covered by anything else. */
     z-index: $zindex-dropdown;
+    /* Animate position together with footer button. */
+    transition: right 0.2s;
```
abias, line 25: Looking at the place where Moodle core sets this style for the footer button on https://github.com/moodle/moodle/blob/master/theme/boost/scss/moodle/layout.scss#L199, you will see that the transition is built with a SCSS include.

So, I think this line should say:

```suggestion
    @include transition(0.2s);
```

abias (changes requested):
Hi @yanus ,

thank you very much for your PR!

It's a shame that I missed that detail myself when I built the back-to-top button.

I had a look at your patch and while it worked smoothly in the browser, I have two change requests regarding the implementation.

As soon as I had the change to fix this, this is ready to be merged.

Cheers,
Alex

abias:
On another note, we are currently discussing the way how to add contributors to the plugin documentation in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/44.

If you want to be mentioned there are code contributor, please tell us your organization / employer (if there is any) and we will add you accordingly.

abias:
Hi @yanus ,

as you allowed the peer reviewer to push to your repo, I just went ahead and applied the small changes to the branch in your repo and will now merge the PR.

Thanks again for your contribution!

abias (approved):

Issue #49, opened 2022-07-08
Feature: Customize activity icon background colors


abias:
Moodle 4.0 introduced a color scheme to group and distinguish the activities in the activity chooser. 

Unfortunately, this color scheme does not work with all corporate identity color schemes. 

The activity colors are defined in SCSS variables in https://github.com/moodle/moodle/blob/master/theme/boost/scss/moodle/variables.scss#L42-L52, so it would be worthwhile to create settings widgets for each color.

abias:
I have submitted a PR for this issue on #51

abias:
This was solved with #51

PR #51, opened 2022-07-11
Feature: Customize activity icon background colors, solves #49

abias:
With this patch, the admin can customize the activity icon background colors which were introduced in Moodle 4.0

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2022-07-08 - Feature: Customize activity icon background colors, solves #49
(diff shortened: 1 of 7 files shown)


abias:
Hi @christianwolters ,

thanks for your peer review!

You are right regarding MOD_PURPOSE_OTHER - this is indeed another activity type. However, if you check the color of this type (for example on https://github.com/moodle/moodle/blob/master/admin/tool/componentlibrary/content/moodle/components/activityicons.md#purpose-colours), you will see that its color is just grey. I chose not to add a color picker for this type as a) it has a neutral color and won't interfere with corporate colors and b) I didn't want to encourage admins to set misleading colors for this type which in fact shouldn't have any color.

Having said that, I will no merge the PR.

Cheers!

Issue #52, opened 2022-07-15
Settings: Divide theme settings into multiple pages


abias:
Currently, this theme's settings are combined within one single tabbed settings page as Boost theme settings are built normally.

However, as we are and will be using many hide_if constraints for our settings, we would run into the
stupid "Too much data passed as arguments to js_call_amd..." debugging message if we would
continue to pack all settings onto just one settings page.

To achieve this goal, we will create a custom admin settings category and fill it with several settings pages.

abias:
This was solved in #53

PR #53, opened 2022-07-15
Settings: Divide theme settings into multiple pages, solves #52

abias:
Currently, this theme's settings are combined within one single tabbed settings page as Boost theme settings are built normally.
However, as we are and will be using many hide_if constraints for our settings, we would run into the
stupid "Too much data passed as arguments to js_call_amd..." debugging message if we would
continue to pack all settings onto just one settings page.
To achieve this goal, we will create a custom admin settings category and fill it with several settings pages.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2022-07-15 - Settings: Divide theme settings into multiple pages, solves #52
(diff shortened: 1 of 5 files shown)


abias:
This is a rather big change for the settings page, but it does not introduce any new functionality. I created a PR for the sake of documentation (instead of just pushing the change to master), but I have merged this myself without peer review as all tests have passed successfully.

PR #54, opened 2022-07-15
Feature: Information banners, helps to resolve #4

### LINKED ISSUE #4: Feature: Information banners
Boost Campus provides an information banner feature. It consists of a perpetual information banner and a time controlled information banner. The documenation can be found on https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus#7-tab-info-banner-settings.

It is requested to include a similar functionality into Boost Union.

Before starting the implementation (or just copying the code from Boost Campus), these additional thoughts should be considered:
* Boost Campus has a hardcoded order of one time controlled banner and one perpetual banner. It might be useful to let the admin control the order of both banners. Similarly, it might be useful to let the admin create more than two banners. See https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus/issues/100 for some more thoughts.
* Boost Campus has a feature to allow Moodle users to dismiss the (perpetual) information banner and allows admins to re-show the banner for all users who dismissed the banner. It might be useful to allow each user to re-show the banner as well, similarly to how users are allowed to re-show user tours.
* The information banner can be especially used to show maintainance period informations to users. It might be useful to provide a CLI script for handling the banners. See https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus/issues/101 for some more thoughts.
* The information banner functionality can get quite huge and might not necessarily have to be a part of a theme. It might be worthwhile to think about outsourcing the information banner feature into a plugin of its own.
* With https://moodle.org/plugins/auth_outage and https://moodle.org/plugins/block_advnotifications, there are two existing implementations for the similar functionality. It might be worthwhile to look at the first to avoid re-inventing the wheel.

abias:
This contribution adds a revamped edition of the information banners from Boost Campus.

The details are listed in #4


christianwolters (dismissed):
## Review

### Documentation

- [x] **Commit Message** understandable and issue referenced
- [x] **Patch author** correct
- [x] **CHANGES.md** appended
- [x] **README.md** appended
- [x] appropriate **Comment quantity**
- [x] successful **Moodle PHPDoc check**

### version.php

- [x] checked if **$plugin->verion** increment necessary (and increment done if necessary)
- [x] **$plugin->release** untouched

### lib.php

- [x] only **necessary functions**

### Languages

- [x] only **english strings**
- [x] necessary **magic strings** added (e.g. capabilities)

### Automated tests

- [x] **new functionality covered**
- [x] **no failing steps or scenarios**

### Github action

- [x] **all green**

### Commit history and scope

- [x] already **single commit**
- [x] focus on **single purpose**
- [x] no **surplus files**

## Review result

**This is a good pull request, ready to merge** 👍

Two remarks:

* I tested the banners on all layouts that are defined in the seetings and the all work as expected. But there may be a need for an additional layout: my courses. In Moodle 3 the course list was part of the dashboard, now it has its own space. Users might be accustomed to having information banners above the course list. We should probably address this in a future release.

* there was a warning issued on the settings tab for the information banners.
This warning may be due to my test setup - maybe not. I seems that functionality is not impacted.

The following warning was displayed:
```
Too much data passed as arguments to js_call_amd("core/showhidesettings", "init"). Generally there are better ways to pass lots of data from PHP to JavaScript, for example via Ajax, data attributes, ... . This warning is triggered if the argument string becomes longer than 1024 characters.

    line 1066 of /lib/outputrequirementslib.php: call to debugging()
    line 173 of /admin/settings.php: call to page_requirements_manager->js_call_amd()
```

abias:
Thank you for your kind peer review, @christianwolters !

Regarding your two comments:

>   I tested the banners on all layouts that are defined in the seetings and the all work as expected. But there may be a need for an additional layout: my courses. In Moodle 3 the course list was part of the dashboard, now it has its own space. > Users might be accustomed to having information banners above the course list. We should probably address this in a future release.

That's a good point! I completely forgot about the MyCourses page.
I will add a commit to cover this page as well before merging the PR.

>     there was a warning issued on the settings tab for the information banners.
    This warning may be due to my test setup - maybe not. I seems that functionality is not impacted.

This warning message was the reason why I split the settings pages in #53, but it couldn't be avoided anyway for the information banners as they are using the hide_if rule in settings.php too often.

The reason for this warning is Moodle core wants to inform you as developer that you shouldn't pass large amounts of data to JS scripts within the JS init method. There are better ways to do that nowadays.

However, the hide_if rule which can be used in admin settings pages is implemented exactly that way. Which means that if you use too many hide_if rules in your settings, you will see this warning. In fact, Moodle core should be changed to use another mechanism for initializing the hide_if rules in the browser. But that's another topic.

PR #62, opened 2022-07-28
Feature: Login page background #36

melanietreitinger:

PR #87, opened 2022-09-26
Changes in config.php fixes #62 #87.

lucaboesch:


abias (approved):
Thank you, Luca, for these fixes!
I am really happy to see these issues fixed.
I think I am the one to blame for these glitches as I did not realize, when building the Boost Union boilerplate, that not all settings of the theme's config.php file are inherited from the parent theme(s).

PR #88, opened 2022-09-27
Improvement: Align the theme's config.php even more with theme_boost

abias:


lucaboesch (approved):
Looking good, @abias !
Merging.

Best,
Luca

PR #92, opened 2022-09-28
stylelint-disable-line declaration-no-important replaced fixes #91.

### LINKED ISSUE #91: Improvement: Fallback CSS file leads to Code prechecker error
In a line (line 13866) of the newly introduced moodle.css fallback file a Code prechecker CSS error "" is triggered due to a wrongly placed /* stylelint-disable-line declaration-no-important */.

lucaboesch:


lucaboesch:
No, the proposed fix isn't probably the right solution.

Issue #75, opened 2022-08-23
Improvement: Month names are always in english


ralf-krause:
Hi,
in the info banner time settings the drop down menu always shows the english month names even I use the language "Deutsch (de)" or "Deutsch Kids (de_kids)". Also the numbers in the other drop down menus should have two digits ... with zero in front if there is only one digit.

Ralf Krause
moodleSCHULE e.V.

<img width="753" alt="Bildschirmfoto 2022-08-23 um 13 22 46" src="https://user-images.githubusercontent.com/27956839/186145910-2c1c723f-4093-4dda-88eb-57c143d27769.png">

abias:
Thank you, Ralf, for reporting this issue.

I created PR #95 to solve this issue.

lucaboesch:
This has been fixed with #94 

<img width="1137" alt="grafik" src="https://user-images.githubusercontent.com/377279/195444614-3b4d3fd6-7719-4641-8f4a-6e349e5fea55.png">

PR #94, opened 2022-09-30
Improvement: Localize month names for time-controlled info banners, solves #75.

abias:


lucaboesch:
Dear @abias this issue reminds me of [https://github.com/catalyst/moodle-MDL-72752/issues/75](https://github.com/catalyst/moodle-MDL-72752/issues/75) we had with Quiz and Question Bank in Moodle 4.1.

And I think it would be more wise to come up using date_selector which should use the language defined date picker [https://docs.moodle.org/dev/lib/formslib.php_Form_Definition#date_selector](https://docs.moodle.org/dev/lib/formslib.php_Form_Definition#date_selector) (so it covers also different element's order e.g. month/day/year as well as localized strings).
What do you think?

Best,
Luca

abias:
Thank you for your wise feedback, @lucaboesch 

The main issue is that Boost Union needs a datepicker in admin settings which does not exist in Moodle core admin libraries yet.
And existing mform widgets can't be used in admin settings without writing an admin setting class as wrapper (to handle the storage, validation and so on).

When I implemented the admin_setting_configdatetime class on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/classes/admin_setting_configdatetime.php to be used by Boost Union, I have to admit that I went the easy way and used an existing solution by Valery Fremaux which turned out to work quite well instead of trying to use a mform widget.

I completely agree that the admin_setting_configdatetime class could be reworked to use the mform date_time_selector widget. However, I would like to split this into a follow-up issue and just fix the localization issues of the existing implementation which have been reported by Ralf with this tiny patch here.

Would you agree?

lucaboesch (approved):
That's fine, @abias 
Thanks for your contribution!

Merging.

Best,
Luca

PR #101, opened 2022-10-12
Re-adding activity navigation fixes #100.

### LINKED ISSUE #100: Restore Previous-Next Activity indicators removed from v4.0-r1 with option to enable/disable them
Re: this forum discussion https://moodle.org/mod/forum/discuss.php?d=439270#p1768055

The activity to activity navigation with the "Previous/Next" buttons and the Jump selector were present in v4.0-r1 but removed from v4.0-r2, because they were removed from core Moodle 4.0 and have yet to be restored despite many user requests (see: https://tracker.moodle.org/browse/MDL-75683 and https://moodle.org/mod/forum/discuss.php?d=433448.)

As suggested by Alexander in the forum reply, it would be good to 1) restore the code back to v4.0-r1 to allow this feature to be used, and 2) add a setting in the theme to enable it, though it would be off by default. Presumably, this should appear in the settings in the Feel > Navigation section along side the "Back to top button" feature. 

As discussed in the forum link above to add this back into core, it is a vital feature for users that need an activity to activity navigation feature. 

boostunion-activitynav

lucaboesch:
I think this should do the trick.

Requesting review.

Best,
Luca

### DIFF
--- a/layout/columns2.php
+++ b/layout/columns2.php
@@ -24,6 +24,7 @@
  * * Render theme_boost_union/columns2 instead of theme_boost/colums2 template
  * * Include course related hints
  * * Include back to top button
+ * * Include activity navigation
  *
  * @package   theme_boost_union
  * @copyright 2022 Luca Bösch, BFH Bern University of Applied Sciences luca.boesch@bfh.ch
@@ -37,6 +38,12 @@ require_once($CFG->libdir . '/behat/lib.php');
 
 // Require own locallib.php.
 require_once($CFG->dirroot . '/theme/boost_union/locallib.php');
+
+// Add activity navigation if the feature is enabled.
+$activitynavigation = get_config('theme_boost_union', 'activitynavigation');
+if ($activitynavigation == THEME_BOOST_UNION_SETTING_SELECT_YES) {
+    $PAGE->theme->usescourseindex = false;
+}
 
 // Add block button in editing mode.
 $addblockbutton = $OUTPUT->addblockbutton();
--- a/tests/behat/theme_boost_union_feelsettings_navigation.feature
+++ b/tests/behat/theme_boost_union_feelsettings_navigation.feature
@@ -83,3 +83,35 @@ Feature: Configuring the theme_boost_union plugin for the "Navigation" tab on th
     # Then The page will be reloaded
     # And The page view will scroll back to x "0" y "250"
     # And The page view will remain at x "0" y "0"
+
+  @javascript
+  Scenario: Setting: Activity navigation - Enable "Activity navigation"
+    Given the following config values are set as admin:
+      | config             | value | plugin            |
+      | activitynavigation | yes   | theme_boost_union |
+    And the following "activities" exist:
+      | activity | name    | course | idnumber |
+      | forum    | Forum 1 | C1     | forum1   |
+      | forum    | Forum 2 | C1     | forum2   |
+      | forum    | Forum 3 | C1     | forum3   |
+    Given I log in as "teacher1"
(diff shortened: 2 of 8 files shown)


tests/behat/theme_boost_union_feelsettings_navigation.feature
```diff
@@ -16,6 +16,11 @@ Feature: Configuring the theme_boost_union plugin for the "Navigation" tab on th
       | user     | course | role           |
       | teacher1 | C1     | editingteacher |
       | student1 | C1     | student        |
+    And the following "activities" exist:
```
abias, line 19: Nice to have:
I would like to propose to move this section out of the background and into each of the "Setting: Activity navigation" scenarios. Doing this, you will have this section twice in this feature file. But you won't burn CPU time uselessly in all other scenarios in this feature file which do not need these activities at all.

layout/columns2.php
```diff
@@ -38,6 +38,13 @@
 // Require own locallib.php.
 require_once($CFG->dirroot . '/theme/boost_union/locallib.php');
 
+$activitynavigation = get_config('theme_boost_union', 'activitynavigation');
```
abias, line 41: I have two minor comments:
* Setting the usescourseindex property here in the layout file seems slightly wrong as the property is already set in the theme's config.php file (see https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/config.php#L49). Of course, you are overwriting the property which is set in config.php here. However, wouldn't it be able to add the logic to set the property to true or false in the theme's config.php only and omit it here?
* If you decide to keep the logic here in the layout file, please
1. move line 41 below the comment in live 43
2. add a modification comment after line 26 to ease future syncs of this file with upstream.

abias (changes requested):
Hi Luca,

many thanks for tackling this issue so quickly.

I just had a look at it and can confirm that in my manual test, the patch did what it's expected to do.

So here's the review checklist:

## Review

### Documentation

- [X] **Commit Message** understandable and issue referenced
- [X] **Patch author** correct
- [X] **CHANGES.md** appended
- [X] **README.md** appended
- [X] appropriate **Comment quantity**
- [X] successful **Moodle PHPDoc check**

### version.php

- [x] checked if **$plugin->verion** increment necessary (and increment done if necessary)
- [X] **$plugin->release** untouched

### lib.php

- [X] only **necessary functions**

### Languages

- [X] only **english strings**

### Automated tests

- [X] **new functionality covered**
- [X] **no failing steps or scenarios**

### Github action

- [X] **all green**

### Commit history and scope

- [X] already **single commit**
- [X] focus on **single purpose**
- [X] no **surplus files**

## Review result

* As you see, you only forgot to bump the version in version.php (as you have introduced a new setting and new language strings). Could you please add that?
* I have added two comments to particular code lines which won't block the integration of this feature but which I would appreciate if you could have a look.

Cheers,
Alex

layout/columns2.php
lucaboesch: Tried that (adding it directly to config.php), it did not work.

tests/behat/theme_boost_union_feelsettings_navigation.feature
lucaboesch: Good point. Done!

layout/columns2.php
abias: All right. Thank you for trying it out!

abias (approved):

abias:
Thank you so much, Luca!

Everything is done now, so I will merge it now!

PR #102, opened 2022-10-12
Adding screenshot image for theme chooser fixes #33.

### LINKED ISSUE #33: Release: Add Screenshot and choosereadme String
Boost Union does not ship with a /pix/screenshot.png file and choosereadme string of its own yet.
Before the first release, this should be added to distinguish Boost Union from Boost.

lucaboesch:
It is hard to come up with an image that shows everything, of course.
For example, there is no way to show a banner in a course page simultaneously to the activity navigation.
I hope this image (a bit faked, e.g. to show the back to top arrow even if on top of a course) distinguishes enough from theme Boost.
It shows the larger content region, the back to top button, a banner, the footer region and the imprint link as well as the modified colors for icons and brand (primary) color.

grafik

Requesting review.


abias:
That's awesome, thank you Luca!

I have just noted that you incremented the version number to today's date. This is not good as we must only increment the last two digits if we want to maintain multiple versions of the plugin in the future (just like Moodle core handles versions).

While incrementing version.php is purely optional for this patch, I fixed this myself during the review and will merge the PR then.

Cheers,
Alex

abias (approved):

Issue #104, opened 2022-10-13
Background image SCSS is included twice


abias:
Unfortunately, the SCSS code for the background images is added twice to the compiled CSS which is shipped to the browser.

![grafik](https://user-images.githubusercontent.com/1092118/195587697-77f7e5ef-77d0-4ff0-8d65-03b4b6362a0e.png)

Issue #103, opened 2022-10-13
Custom SCSS is included twice


abias:
Unfortunately, the custom SCSS which is added in the Boost Union settings, is added twice to the compiled CSS which is shipped to the browser.

![grafik](https://user-images.githubusercontent.com/1092118/195584665-06a25751-fad8-4232-8d19-33bcac401347.png)

PR #105, opened 2022-10-13
Bugfix: Custom SCSS and background image SCSS was included twice, solves #103 #104

abias:


lucaboesch (approved):
Looking good, @abias!

Merging.

Issue #107, opened 2022-10-14
Login page background is applied to the #page element instead of the body element


abias:
The login page background (which is a Boost core feature) is added to the #page element instead of the body element. This is architecturally strange as the normal background image is added to the body element indeed. Additionally, as soon as a footnote is added in Boost Union, the login page background does not cover the whole page anymore and the normal background image (which is added to the login page as well by Moodle core) is shining through:

![grafik](https://user-images.githubusercontent.com/1092118/195770325-7a18b21e-e5da-4800-b07f-928e62663da9.png)

Issue #106, opened 2022-10-13
Footnote does not have a white background


abias:
As soon as you add a background image to Boost Union, you will notice that the footnote does not have a white background and its content might become unreadable.

![grafik](https://user-images.githubusercontent.com/1092118/195595304-717085fd-5651-4b78-aa86-65c0b9f26e9f.png)

PR #108, opened 2022-10-14
Footnote and Login background - Bugfixes for #106 and #107

abias:
#106 - This patch solves the problem that the footnote content was not really readable as soon as a background image was configured. Additionally, it improves the margins and paddings of the footnote on the various screen sizes.

#107 - This patch changes the way how the login page background is added to the page from the #page element to the body element.
Additionally, it introduces several adjustments to make sure that the footnote works well with background images on the login page.

Here's how it looks like now:

# Course page without any background images configured

## Large screen
grafik

## Smaller screen
grafik

## Really small screen
grafik

# Course page with a background image configured

## Large screen
grafik

## Smaller screen (which does not have a background image as Boost Core built it)
grafik

## Really small screen (which does not have a background image as Boost Core built it)
grafik

# Login page with a normal background image configured

## Large screen
grafik

## Smaller screen
grafik

## Really small screen
grafik

# Login page with a login background image configured

## Large screen
grafik

## Smaller screen
grafik

## Really small screen
grafik

# Login page with both login background image and normal background image configured

## Large screen
grafik

## Smaller screen
grafik

## Really small screen
grafik

# Login page without any background images configured

## Large screen
grafik

## Smaller screen
grafik

## Really small screen
grafik

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -164,6 +164,38 @@ function theme_boost_union_get_extra_scss($theme) {
     // We have to accept this fact here and must not copy the code from theme_boost_get_extra_scss into this function.
     // Instead, we must only add additionally CSS code which is based on any Boost Union-only functionality.
 
+    // In contrast to Boost core, Boost Union should add the login page background to the body element as well.
+    // That's why we first have to revert the background which is set to #page on the login page by Boost core already as soon as a
+    // login background image is set. Doing this, we also have to make the background of the #page element transparent on the login
+    // page.
+    $loginbackgroundimageurl = $theme->setting_file_url('loginbackgroundimage', 'loginbackgroundimage');
+    if (!empty($loginbackgroundimageurl)) {
+        $content .= 'body.pagelayout-login #page { ';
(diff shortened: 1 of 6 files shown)


lib.php
```diff
@@ -164,6 +164,38 @@ function theme_boost_union_get_extra_scss($theme) {
     // We have to accept this fact here and must not copy the code from theme_boost_get_extra_scss into this function.
     // Instead, we must only add additionally CSS code which is based on any Boost Union-only functionality.
 
+    // In contrast to Boost core, Boost Union should add the login page background to the body element as well.
+    // That's why we first have to revert the background which is set to #page on the login page by Boost Core already as soon as a
```
lucaboesch, line 168: Please write Core in Boost core in lower case consistently.

abias: Thanks, Luca, for this hint! This is changed now.

lucaboesch (approved):
Those changes are looking good.
Thanks, @abias!

Issue #98, opened 2022-10-11
Improvement: Further hints on hidden courses


Menrath:
#5 is very helpful! In the past even though we had a similar feature (Show hint in hidden courses) in our theme, some teachers still believed that creating a new topic in the notifications forum would trigger (email) notifications. It might seem nitpicky, but adding a warning automatically via `showhintcoursehidden` also on the forum page might be of use from time to time. Pull Request would be kind of ready, as I already played around with it.

lucaboesch:
As far as I can tell, nothing can be said against a pull request!
At max, you get some feedback on how to improve or why that woudn't fit.
Be sure to cover it by a settings toggle to enable/disable this feature.
If you are a wizard you might also cover your feature by Behat tests, but don't worry if that is not your thing.
Just pull request away.

Best,
Luca

andremenrath:
Thanks for your quick response!

As one could add an extra setting for that, at least for me it makes sense to bind it to the `showhintcoursehidden` setting. Don't you agree? But maybe it's better if I just share my draft in a PR and let's discuss it over there.

Greetings, André

goggo24:
Hey, which version do you prefer:
a seperate aetting or include it into showhintcoursehidden? Depending on your answer we will create the PR. 
Thanks! Eva

wiebkemueller-hsh:
Sounds very useful! As a user I'd be thankful for that feature and it would be ok to activate it within the showhintcoursehidden. 

The description could be changed to 

Hinweis**_e_** in verborgenen Kursen anzeigen

"Mit dieser Einstellung wird ein Hinweis im Kursheader **_und in der Aktivität Forum Ankündigungen_** angezeigt, solange der Kurs verborgen ist. Dies erlaubt die Erkennung des Sichtbarkeitsstatuses eines Kurses auf einen Blick, ohne die Kurseinstellungen öffnen zu müssen."

Admin-wise, what do you think @abias ?

Regards from H
Wiebke

abias:
Hi @Menrath ,

looking at our [project board](https://github.com/orgs/moodle-an-hochschulen/projects/4), I just spotted this issue which has been in "In progress DEV" status for quite a while now, but the discussion here and in the corresponding PR #109 somehow got stale and you have closed the PR several weeks ago as well.

I would like to encourage you to rebase and revise the code which you submitted in #109 and just control the 
[...]

PR #109, opened 2022-10-14
Additional nitpicky hidden course hint on forum pages

andremenrath:
See Issue https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/98

This PR currently copies the logic of the hiddencoursehint and uses a seperate similar mustache template.
I thought of using the same template file with an additional templatecontext to choose which string to use.

It currently provides no additional setting for controlling this additional but related hint, which has to/may be discussed.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -40,7 +40,7 @@ function theme_boost_union_get_course_related_hints() {
     // Initialize HTML code.
     $html = '';
 
-    // If the setting showhintcoursehidden is set and the visibility of the course is hidden and
+    // If the setting showhintcoursehidden is set and the visibility of the course is hidden
     // a hint for the visibility will be shown.
     if (get_config('theme_boost_union', 'showhintcoursehidden') == THEME_BOOST_UNION_SETTING_SELECT_YES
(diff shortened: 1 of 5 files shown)


locallib.php
```diff
@@ -62,6 +62,28 @@ function theme_boost_union_get_course_related_hints() {
         $html .= $OUTPUT->render_from_template('theme_boost_union/course-hint-hidden', $templatecontext);
     }
 
+    // If the setting showhintcoursehidden is set and the visibility of the course is hidden and
```
lucaboesch, line 65: The ‚ and‘ at the end of the line is almost surely superfluous.

lucaboesch:
It would be great if that got a separate admin side setting to set if that hint is shown in forums with Yes and No as options.
Default should be No. We want out-of-the box Boost Union be looking like unmodified boost.
Thanks for your hard work!

Menrath:
@lucaboesch I would still propose to bind this additional feature to the existing settings which controls the hidden course hint. I totally agree with the out-of-the-box guidelines but I would not like to see the admin-settings page getting bloated on very similar features, where it is unlikely admins would choose only one option off the two. What do you think?

Issue #113, opened 2022-10-15
Feature: Filearea for additional resources


abias:
Boost Campus provided a custom filearea for additional resources. These additional resources could be referenced from anywhere on any page and could especially be used to add images to custom SCSS.

It would be worthwhile to port this feature over to Boost Union.

PR #115, opened 2022-10-18
Feature: Filearea for additional resources, solves #113

abias:
With this patch, a setting is added where the admin can upload additional resources to the theme which can be linked from anywhere in Moodle (especially from custom CSS).


lucaboesch (approved):
These contribution is looking good!
Merging.

Issue #38, opened 2022-07-01
Feature: Filearea for custom fonts


abias:
Boost Campus provided a custom filearea for custom font files. These font files could be referenced in custom SCSS later.

It would be worthwhile to port this feature over to Boost Union.

abias:
Beyond just adding a dedicated file area for font files, it would be awesome if the theme could automatically parse the uploaded fonts (with a library like https://github.com/dompdf/php-font-lib) and automatically add the fonts to the CSS for direct usage.

ralf-krause:
The filearea for font uploads is very important because it‘s a privacy issue to embed fonts from Google server on the fly.
https://www.ra-plutte.de/lg-muenchen-dynamische-einbindung-google-web-fonts-ist-dsgvo/

ralf-krause:
News about using Google fonts
https://www.heise.de/news/DSGVO-Abmahnwelle-wegen-Google-Fonts-7206364.html

abias:
Based on the work in #113 , I have implemented a custom fonts filearea on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/tree/issue-38
Currently, it does not provide any real additional value beyond the additional resources filearea from #113.
This is because the fonts are not parsed yet and the admin still has to compose the CSS snippets manually. 

I had a quick look at https://github.com/dompdf/php-font-lib which looked promising at the beginning, but does not provide WOFF2 parsing (which is a must-have from my point of view) yet.

@lucaboesch , I would be grateful if you would add your two cents to this to see if it would be useful to integrate this feature already or to wait for a working fonts parsing solution.

If it will be integrated as-is, some quick Behat tests and a version bump will be needed, but nothing more.

lucaboesch:
@abias I think it is fully valid to include this although there is no fonts parsing solution yet.

abias:
Allright, Luca, here it is: https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/116
And I created a follow-up issue on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/117

lucaboesch:
Solved by #116

PR #116, opened 2022-10-19
Feature: Filearea for custom fonts, helps to solve #38.

abias:
With this setting the admin can upload custom fonts to the theme.


lucaboesch (approved):
This is looking good to me, @abias.
Merging.

Issue #120, opened 2022-10-26
Improvement: Restrict uploadable filetypes in custom fonts filearea


abias:
During the implementation of #116, it was missed to restrict the set of uploadable filetypes to webfonts extensions. To avoid that arbitrary files are uploaded here, this restriction should be added.

lucaboesch:
This is done and can be closed.

PR #121, opened 2022-10-26
Improvement: Restrict filetypes in custom fonts filearea, solves #120.

abias:


locallib.php
```diff
@@ -692,3 +692,60 @@ function theme_boost_union_get_customfonts_templatecontext() {
 
     return $filesforcontext;
 }
+
+/*
```
lucaboesch, line 696: Put two stars here
/ **

locallib.php
```diff
@@ -692,3 +692,60 @@ function theme_boost_union_get_customfonts_templatecontext() {
...
+ * Helper function which makes sure that all webfont file types are registered in the system.
+ * The webfont filetypes need to be registered in the system, otherwise the admin settings filepicker wouldn't allow restricting
+ * the uploadable filetypes to webfonts only. *
```
lucaboesch, line 699: Remove the star at the end of the line.

lucaboesch, line 699:
```
 * @return void
 * @throws coding_exception

```
do help here (not sure about the @returns but the @throws has to be)

locallib.php
```diff
@@ -692,3 +692,60 @@ function theme_boost_union_get_customfonts_templatecontext() {
...
+ */
+function theme_boost_union_register_webfonts_filetypes() {
+    // Our array of webfonts to register.
```
lucaboesch, line 702: Better would be:
`    // Our array of webfonts filetypes to register.`

abias:
Thank you, Luca, for your quick review.

I managed to fix all raised items and pushed another version.

lucaboesch (commented):
That's good now.
Merging.

lucaboesch (approved):

Issue #133, opened 2022-11-09
Bug: Boost Union blocks Moodle 4.0.4+ (20221108)


ralf-krause:
I got an error when I updated my MoodleBox up to Moodle 4.0.4+ (20221108).

```
Coding error detected, it must be fixed by a programmer: 
Cannot set custom filetypes because they are defined in config.php
```

First I thought that the error is specific for the MoodleBox https://moodlebox.net . I updated two other platforms to the same Moodle version. On my MoodleBox it was not possible to open the web administration anymore.

After I changed the parameter customfiletypes to xcustomfiletypes I could open the web administration again. Yes, I know that it does not make sense to change a parameter name. I did this to find a solution for the problem. In the MoodleBox config.php there is a part which defines a custom file type.

```
$CFG->customfiletypes = array(
  (object)array(
    'extension' => 'crt',
    'icon' => 'sourcecode',
    'type' => 'application/x-x509-ca-cert',
    'customdescription' => 'X.509 CA certificate'
  )
);
```

Nicolas Martignoni who is the developer of the MoodleBox did exactly what he should do to define a custom file type. Please look into this [comment in config-dist.php](https://github.com/moodle/moodle/blob/5877ae306efaa93a3a0cf8adcedee4da405677f5/config-dist.php#L613-L628).
You also can read https://github.com/moodlebox/moodlebox/issues/273

Okay, I should create a tracker issue for this because the problem is the result of the Moodle update from 4.0.4+ (20221104) to 4.0.4+ (20221108).
https://tracker.moodle.org/browse/MDL-76247

No, the problem doesn't come from the Moodle code. Jake Dallimore used that exact configuration and had been able to upgrade just fine to 4.0.4 weekly. He was closing this one as cannot reproduce.

So I took my own distribution Moodle4mac 4.0.4 from https://download.moodle.org/macosx/ ... I just updated to Moodle 4.0.4+ (20221104) last weekend. So I had a good start for testing. I added the customfiletypes definition from the MoodleBox to the config.php. Everything run fine. I updated to Moodle 4.0.4+ (20221108) now. And Moodle runs without any problem.

Now I installed the Boost Union v4.0-r6 theme_boost_union_moodle40_2022080910
Yes, just in this moment the problem came up. No other additional plugin is installed.

[...]

PR #134, opened 2022-11-09
Bugfix: Site admin broken if customfiletypes set, solves #133

abias:


abias:
I implemented a fix for #133.
As this is a major issue with the latest release and, on the other hand, the fix is quite straightforward, I will merge this without a proper peer review to be able to push out a fixed release as soon as possible.

Issue #59, opened 2022-07-27
Feature: Add FontAwesome 5 or 6 to the theme


abias:
Moodle core still ships with FontAwesome 4 which is really outdated. The discussion to upgrade FontAwesome to version 5 in https://tracker.moodle.org/browse/MDL-64853 was closed as "won't fix" and another discussion to move to ForkAwesome has been started in https://tracker.moodle.org/browse/MDL-67448.

Regardless of these core discussions, it would be awesome to be able to use FontAwesome 5 icons on sites with Boost Union out of the box, especially with the https://moodle.org/plugins/filter_fontawesome filter.

It would be worthwhile to check if it is technically possible and, regarding the FA license allowed, to add FontAwesome 5 Free to Boost Union. Additionally, it would be great if there would be a possibility to use the Pro icons as well if there is a valid FontAwesome license.

ralf-krause:
Does this link help to answer your question? It seems to be allowed to host the FontAwesome 5 Free or FontAwesome 6 Free icons in Boost Union. For compatibility to the FontAwesome 4 icons in Moodle core you need to add some code. `In addition to whatever v6 CSS files you're including, you'll want to include the following v4-shims.js file, which will map v4 icon names to icons in the latest version of Font Awesome.`

ralf-krause:
Moodle closed the https://tracker.moodle.org/browse/MDL-64853 to implement ForkAwesome. In the FontAwesome 5 Free and FontAwesome 6 Free version there are missing a lot of icons.

abias:
This has been solved by #141 now

PR #141, opened 2022-11-15
Feature: Possibility to upload FA 6 Free to the theme, solves #59

abias:
This patch introduces a new setting in the Look -> Resources tab which allows the admin to upload FontAwesome 6 Free files and add them to Moodle.

grafik

As soon as some files are uploaded, a file list is shown which helps the admin to identify the remaining files and which tells him which files are mandatory and which are optional:
grafik

Finally, there is a checks section which helps the admin to assess if everything is working correctly:
grafik

During the implementation of this patch, some decisions have been made:
* Currently, only FontAwesome 6 Free is supported. However, the architecture is already built in a way to allow the addition of FontAwesome 6 Pro quite easily. I will create a follow-up issue for that.
* Currently, FontAwesome 5 is not supported. It seemed not to make much sense to allow outdated FA versions as well if the purpush of this feature is to allow admins to use the latest icons.
* Currently, the FontAwesome 6 fonts are added to the Moodle page _in addition_ to the FontAwesome 4 fonts from Moodle core. Both reside in their own font family namespaces and I did not see any clashes in my tests. However, I will create a follow-up issue to clean this up if possible.
* Currently, adding a full set of Behat tests is not possible with Moodle core steps. I will create a follow-up issue for that as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2022-11-15 - Feature: Possibility to upload FontAwesome 6 Free to the theme, solves #59.
 * 2022-11-08 - Feature: Allow admins to override the email templates within the theme, solves #60.
 
 ### v4.0-r7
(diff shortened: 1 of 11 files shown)


lucaboesch (approved):
I approve of this changes and find them good.
Merging.

ksteitz:
**Font Awesome 6 implemented since Moodle 4.2**
Since Moodle 4.2 FA 6.3 was implemented after all with and thus the change from 4 to 6: https://tracker.moodle.org/browse/MDL-76989 (Upgrade Font Awesome to 6.3.0), lib/fonts/    4.7.0  -->  6.3.0
Is that FA union-feature therefore redundant starting with v4.2-r1? Or at least needs to be modified, leaving an option to manually add (pro)-icons? 
If it helps i can create a new issue for this. I am not aware of the correct issue-policy.

Issue #147, opened 2022-11-19
Improvement: Restrict accepted file types for background images


abias:
Currently, the background image setting (which is replicated from theme_boost) accepts arbitrary files types.
We should restrict it to the web_image typeset like we already do for the login background images

abias:
This was solved by #148

PR #148, opened 2022-11-21
Improvement: Restrict file types for background images, solves #147

abias:


abias:
As this is a rather easy change and as a new release is scheduled for today, I dared to merge this myself.

PR #179, opened 2023-01-05
Bugfix: Unparsable mustache JSON content.

lucaboesch:
This has slipped us.


abias (approved):
Oh, I really missed that.
Thank you @lucaboesch for spotting and fixing this!

Issue #181, opened 2023-01-08
Bug: Infobanners are sometimes incorrectly ordered if the same order is given to multiple banners


abias:
During the review of #167, I noticed that infobanners are sometimes incorrectly ordered if the same order is given to multiple banners and the current tests do not detect these cases.

We have to change the implementation on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/locallib.php#L414-L419 to a more robust one which includes the `no` attribute of the banner into the comparison instead of relying on some internal behaviour of usort.

abias:
This was solved by #182

PR #182, opened 2023-01-08
Bugfix: Infobanners were sometimes incorrectly ordered, solves #181

abias:


abias:
I will merge this PR without a formal peer review as the fix is a no-brainer and the fix is blocking me from finishing #167

PR #189, opened 2023-01-14
Improvement: Static pages headings and links multilang.

lucaboesch:


locallib.php
```diff
@@ -300,7 +300,8 @@ function theme_boost_union_get_staticpage_link($page) {
  */
 function theme_boost_union_get_staticpage_pagetitle($page) {
     // Get the configured page title.
-    $pagetitleconfig = get_config('theme_boost_union', $page.'pagetitle');
+    $pagetitleconfig = format_string(get_config('theme_boost_union', $page.'pagetitle'), false,
+    ['context' => \context_system::instance()]);
```
abias, line 304: It took me some seconds to understand that using the system context is correct here - we do not want to break the multilanguage links to the static pages in the footer of a course page just because a teacher disabled the multilang filter in his course.

Maybe you want to add this a comment to the code?

locallib.php
```diff
@@ -300,7 +300,8 @@ function theme_boost_union_get_staticpage_link($page) {
  */
 function theme_boost_union_get_staticpage_pagetitle($page) {
     // Get the configured page title.
-    $pagetitleconfig = get_config('theme_boost_union', $page.'pagetitle');
+    $pagetitleconfig = format_string(get_config('theme_boost_union', $page.'pagetitle'), false,
```
abias, line 303: Why are you setting the second parameter (striplinks) from the default (true) to false? The pagetitle configuration does not contain any links at all, the link will be composed later in mustache from the pagetitle and the page's URL.
In my opinion, there is no need to let any links (which the admin might have set in the pagetitle configuration by mistake or maliciously) pass through.

tests/behat/theme_boost_union_contentsettings_staticpages.feature
```diff
@@ -190,25 +190,33 @@ Feature: Configuring the theme_boost_union plugin for the "Static pages" tab on
...
   Scenario Outline: Setting: Static page page title - Set a custom static page page title
     Given the following config values are set as admin:
-      | config             | value              | plugin            |
-      | enable<page>       | yes                | theme_boost_union |
-      | <page>content      | <p>Lorem ipsum</p> | theme_boost_union |
-      | <page>pagetitle    | Custom             | theme_boost_union |
+      | config          | value                                                                                             | plugin            |
+      | enable<page>    | yes                                                                                               | theme_boost_union |
+      | <page>content   | <p>Lorem ipsum</p>                                                                                | theme_boost_union |
+      | <page>pagetitle | <span lang="en" class="multilang">Custom</span><span lang="de" class="multilang">Angepasst</span> | theme_boost_union |
```
abias, line 196: While this enhanced test is perfectly fine, I feel tempted to leave the existing test as it is and to move the multilanguage feature into a dedicated test. In this case, you could also add test steps for configuring multilanguage static page _content_.

What do you think?

abias (changes requested):
Thank you @lucaboesch for handling this improvement so instantly!

I just tested the code and it worked perfectly fine for me. 
However, I have added three minor comments to the code which you might want to handle if you have time before we merge this.

Additionally, you might want to add a line to CHANGES.md about the change. Just to make the PR complete.

Cheers,
Alex

locallib.php
lucaboesch: Pardon me, @abias. I hat the CHANGES.md ready but forget to add it to the commit. And I do always interpret striplinks the wrong way around.

tests/behat/theme_boost_union_contentsettings_staticpages.feature
lucaboesch: If you follow the tradition of Moodle core, those multilang tests are in the existing ones.

lucaboesch:
Pardon me, @abias. I had the CHANGES.md ready but forget to add it to the commit. And I do always interpret striplinks the wrong way around.

tests/behat/theme_boost_union_contentsettings_staticpages.feature
lucaboesch: So, if test steps for configuring multilanguage static page content are to be included in the existing ones, fine. But not separate ones please.

locallib.php
abias: The comment-to-code ratio is quite high in our plugins and comments do not cost anything. But well, it's your patch. Let keep it as it is.

tests/behat/theme_boost_union_contentsettings_staticpages.feature
abias: Ah, I was not aware that Moodle core builds multilang tests that way. In this case, I'm with you.

Regarding static page content, I will check if I can sneak some additional steps in in a separate PR.

abias (approved):

abias:
Everything is solved, let's merge it :)

Issue #191, opened 2023-01-15
Improvement: Add note about charsets in filenames of uploaded asset files


abias:
This is a outsourced issue of Luca's comment in https://github.com/melanietreitinger/moodle-theme_boost_union/commit/fbba83c03bc7aa738fdfc95ab42aa9fed6fb3ce9#r96186790

In Boost Union, admins can upload files in multiple locations which are then shipped by the webserver within the page. For example: Background images and logos, but also assets like font files.

It might be worthwhile to add a note to each of these settings that the admin should not use non-ASCII-characters in the filenames if he wants to be sure that they are served properly by the webserver.

Note by @lucaboesch : To me it seems that the only thing that would break is the matching between login background image and the desired login background image caption

abias:
This was solved by @lucaboesch in #193

PR #193, opened 2023-01-15
Improvement: Matching login image/login image text note, solves #191.

lucaboesch:


CHANGES.md
```diff
@@ -6,6 +6,8 @@ Changes
 
 ### Unreleased
 
+* 2023-01-15 - Improvement: Note to assure matching login image/login image text.
+* 2023-01-14 - Improvement: Add Behat test steps for multilanguage static page content.
```
abias, line 10: This line is not related to this PR and is already there two lines further down in this file. Maybe you didn't apply this commit onto the latest master?

abias (changes requested):
Hi Luca,

thank you for adding this improvement.

I also saw your note on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/191#issue-1533702403 and basically agree that the note about non-ascii characters is mainly important here and uploaded files should be served by the webserver even with special characters in the filename. Additionally, an experienced admin should be aware that filenames should only contain ascii characters.

There is just one minor glitch in your patch in CHANGES.md.
I will go ahead and fix that in your branch so that the PR can be merged.

lang/en/theme_boost_union.php
```diff
@@ -100,7 +100,8 @@
 $string['loginbackgroundimagesheading'] = 'Login page background images';
 // ... ... Setting: Login page background image.
 $string['loginbackgroundimage'] = 'Login page background images';
-$string['loginbackgroundimage_desc'] = 'The images to display as a background of the login page. One of these images will be picked randomly and shown when the user visits the login page.';
+$string['loginbackgroundimage_desc'] = 'The images to display as a background of the login page. One of these images will be picked randomly and shown when the user visits the login page.
+Please make sure not to use non-ASCII-characters in the filename if you want to display text for login background images.';
```
abias, line 104: You have added a line break within a string. This does not directly result in a line break on the settings page. You have to include a `<br>` tag into the string to trigger a line break. 
However, I would propose to remove the line break entirely and just append to new string to the existing string.

abias (changes requested):

Issue #192, opened 2023-01-15
Improvement: Trim background image text color setting


abias:
This is a outsourced issue of Luca's comment in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/69#issuecomment-1382850886.

There he wrote:

> in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/locallib.php#L652 please do trim
> 
>                `if (trim($settings[2]) != 'dark' && trim($settings[2]) != 'light') {`
> because I saw strings ending with \r and they would be rephrased to dark even when saying light\r.

abias:
I just double-checked that the login background text setting is the only place where we are exploding settings like that up to now in Boost Union.

However, it might make sense not only to trim `$settings[2]`, but also `$settings[1]` and `$settings[0]`

abias:
This has been solved in #194

PR #194, opened 2023-01-15
Feature: Trim dark and light for login bg images, solves #192.

lucaboesch:


locallib.php
```diff
@@ -647,14 +647,14 @@ function theme_boost_union_get_loginbackgroundimage_text() {
...
+            if (strcmp($filename, trim($settings[0])) == 0) {
                 // If the color value is not acceptable, replace it with dark.
-                if ($settings[2] != 'dark' && $settings[2] != 'light') {
+                if (trim($settings[2]) != 'dark' && trim($settings[2]) != 'light') {
                     $settings[2] = 'dark';
                 }
 
                 // Return the text + text color that belongs to the randomly selected image.
-                return array(format_string($settings[1]), $settings[2]);
+                return array(format_string(trim($settings[1])), trim($settings[2]));
```
abias, line 657: You are trimming the same string three times in a row. This is not really efficient. You should do it once and store the result back to the array. 

I will change that myself before merging the PR.

abias (changes requested):
Thank you for working on this, Luca!

I found two things which should be changed before merging this PR. Both are minor things, but we strive for clean solutions, don't we?

lucaboesch:
Yes, sure, @abias I agree fully with both (the amendments and that we go for clean solutions). Please proceed as proposed.

abias (approved):

abias:
The tests have passed again, so I will merge this now.

Issue #201, opened 2023-01-22
Feature: Allow the admin to adapt the H5P width


abias:
In scss/moodle/variables.scss, there is the variable `$h5p-content-maxwidth: 960px !default;`.
Similar to the variable `$course-content-maxwidth` which can already be customized in Boost Union, it would be great to allow the admin to customize the H5P width

ralf-krause:
Great idea! 
I use it in some of my administrated platforms.

lucaboesch:
Since #202 has merely fixed the H5P width within content bank, I would pledge to leave this open.

abias:
@lucaboesch , I basically agree with you that this issue is not fully solved yet. However, it was just created as I spotted the existing SCSS variable and wanted to add a setting for it. I do not feel a reel need to control the H5P activity width separately from all other activities.

I rather think that the remaining aspect of this issue is now a subset of #18 and could be handled there. What do you think?

abias:
I dare to close this now based on my comment above.

Issue #18, opened 2022-03-22
Proposal: Streamlining of content width sizing


ShUl0w:
In its current form, the new Moodle 4.0 UI uses an inconsistent approach to content sizing for center column content. This is especially notable on pages where the center column alternates between a **.container** for its course related contents (e.g. course navigation), and a **.container-fluid** for elements that contain tables etc. (e.g. gradebook).  

This becomes even more of a problem when modals are involved, as they seem to always use a .**container**, which in turn loses precious screen real estate; the new question bank UI is a good example of this.

Given how the 4.0 UI is already trying to use the full screen width by including a drawer on the left and right side of the screen respectively, I propose to streamline the content width across all screens if possible and simultaneously increase the size of the center column. Adapting the drawers to reduce the middle column size when visible shouldn't be that much of a problem.

dbnschools:
$course-content-maxwidth: 100% !default;
makes the pages all full-width.  Drop it in the Raw SCSS box of Boost.

abias:
In #45, I just submitted a PR which allows admins to overwrite the $course-content-maxwidth SCSS variable in the GUI. This can help to resolve this issue, but I assume that it is not sufficient at all.

abias:
#18 closed this automatically. 
I reopen this issue for further discussions

ralf-krause:
Yes, I like the feature for setting the course width to 1200px for large displays.
But it would also be nice to reduce the borders on the right and the left side for small displays as we have with smartphones or tablets. I tested this in the raw scss section.

```
/* Seitenränder für kleine Displays reduzieren */
@media (max-width: 767.98px) { 
#page.drawers, #page.drawers div[role="main"] { padding: 0 3px 0 3px; } 
}
@media (min-width: 768px) { 
#page.drawers { padding: 0 10px 0 10px; } 
}
```

wiebkemueller-hsh:
As discussed with @ShUl0w this can be closed!

PR #202, opened 2023-01-22
Feature: Allow the admin to change the H5P content bank width, solves #201

abias:


abias:
After @lucaboesch 's (private) feedback for me, I realized that this SCSS variable only applies to the H5P content in the content bank, but does not apply to H5P content within the H5P activity. That's sad but true. 
However, I have modified the setting descriptions so we can still merge this feature in - although it is only a minor feature anoymore. 

In #18, we can still discuss further improvements on content width.

lucaboesch (approved):
This looks good.

lucaboesch:
This looks good.
Merging.

Issue #211, opened 2023-01-28
Feature: Add dedicated logo upload to Boost Union


abias:
The logo and compact logo upload in Moodle core is limited - it only accepts JPG and PNG images. To be able to solve #160 and to be able to further harmonize the flavour logo with the global logo, Boost Union should get its own logo upload.

abias:
This was solved by #213

Issue #160, opened 2022-12-04
Feature: Allow SVG files for the logo and compact logo setting


abias:
The logo and compact logo setting is located in the Moodle core, not in Boost Union. Boost Union just shows the logo from the Moodle core configuration. Unfortunately, Moodle core restricts the logo file formats to PNG and JPG. There is a discussion about removing the limitations in https://tracker.moodle.org/browse/MDL-61963, but this got stuck.

It would be worthwhile to add the possibility to upload SVG files as well and make sure that they are properly displayed as logo.
However, this would come with the requirement to duplicate the logo setting into Boost Union and the ignore the logo which is uploaded in the global settings.

abias:
In #213, I have made some basic steps towards supporting SVG logos. In Chrome, it seems to work quite fine already:

<img width="1440" alt="grafik" src="https://user-images.githubusercontent.com/1092118/215293813-14a4ba5a-33a5-4a64-b54a-f859cc166a15.png">

However, Firefox does not get the width of the SVG image in the compact logo setting yet and overlaps the logo with the navbar links:

<img width="1440" alt="grafik" src="https://user-images.githubusercontent.com/1092118/215293832-b212a502-47f7-467b-8306-11e8636fc09c.png">

Some more investiation will be needed.

abias:
The remaining issue with SVG logos in Firefox was solved in #221 221

PR #213, opened 2023-01-28
Restructurings for Boost Union's logo handling

abias:
This PR solves #211 which should land before the 4.1 upgrade (to make the upgrade more understandable as from 4.1 on, we will also have duplicated favicon settings).

It also tries to solve #160 and add SVG logo support, however this won't be production ready yet and will need some more tweaks and testing later.


lucaboesch (approved):
This is all looking good.
Thanks, @abias.

lucaboesch:
Merging.

Issue #218, opened 2023-02-05
Improvement: Add more horizontal margins to the infobanners


abias:
Currently, the infobanners do not have the same horizontal margins as the rest of the Moodle content.
This becomes obvious and starts to look odd as soon as you add a background image to the site:

<img width="1440" alt="grafik" src="https://user-images.githubusercontent.com/1092118/216806083-e7881ff1-22a4-41fc-9847-de89ee9d430d.png">

<img width="1440" alt="grafik" src="https://user-images.githubusercontent.com/1092118/216806093-7cbc05fd-d95f-420c-a23b-1baa9d8a92de.png">

We should align the horizontal margin with the main content.

abias:
This was solved in #222

PR #222, opened 2023-02-10
Fix: Align horizontal margins for the info banners.

andremenrath:
Solves issue #218

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


abias (approved):
Hey @andremenrath ,

many thanks for this bugfix! 

From my point of view, your solution is spot-on and worked in my tests.
I just added an additional commit on top of your branch which adds a small suffix to the SCSS comment and adds an entry to the CHANGES.md file.

I will merge the PR as soon as the tests have passed.
I addition to that, I will add you to the list of Boost Union contributors in README.de in a second step.

Cheers,
Alex

Issue #39, opened 2022-07-01
Feature: Dark navbar


abias:
Boost Campus provided the possibility to invert the navbar and make it dark.

It would be worthwhile to port this feature over to Boost Union.

abias:
@t-schroeder - As discussed in the HU Forums, I have added you as assignee to this issue.

Amrita1991:
Hi,

I have created a pull request for this feature
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/63

lucaboesch:
I heard the request of having not just dark but a configurable color.

andremenrath:
@Amrita1991 @lucaboesch @t-schroeder 
Is anyone already putting work into this (Not only dark, but branding-color styled navbar)? We at AMC would like to see this future and could work on this. Should we use a different issue for it?

lucaboesch:
I from my point can say: go ahead, @andremenrath !
What about you, @Amrita1991 and @t-schroeder ?

t-schroeder:
@lucaboesch What should be the default color if the color is made configurable?

@andremenrath We were not yet planning to work on this. So if you want to implement this, that's fine with me. Making the color configurable would obsolete the dark navbar setting. The proper way would probably be to close this issue and create a new one.

abias:
As @Amrita1991 was the first one to submit a PR to this issue, I have changed the assignee of this issue.
@Amrita1991 's PR was finally reviewed any is now superceeded by #224 which will hopefully be merged soon and which will then solve this issue.

abias:
As #224 was merged just now, this issue is solved now.

PR #224, opened 2023-02-12
Feature: Allow admin to change the navbar color, solves #39, helps to resolve #110

### LINKED ISSUE #110: Feature: Custom color-branded navbar
Extends #39 

I started work on this but stumbled into some problems, some can be resolved others have some further impact.

**Resolved:**
Automatically choose the right font-color depending on the contrast, etc.

**Needs discussion:**
The navbar background color might interfere with the `brandcolor` therefore the navigation highlighting of the currently active tab and the edit button might not be visible good enough. I am not sure anymore how much this setting can be generalized in a meaningful way.

abias:
This replaces the PR #84 which was reviewed by me but where I decided to put some more effort in it and extend the contributed functionality.

It allows the admin to switch to a dark navbar:
grafik

and it also allows the admin to switch to a navbar which is colored with the brand color:
grafik

I will ask my designer colleagues if they can tweak the dark and colored navbar even more and submit the tweaks as PR, but this is just a matter of SCSS improvements and won't change the functionality.


lucaboesch (dismissed):
Unfortunately, it is not that beautiful when you have the same identical string twice in the settings, and once where you would expect additional explanations.
image
Please refine this patch regarding this admittedly very small detail which is visible to admins only anyway.

settings.php
```diff
@@ -389,6 +389,29 @@
...
+        // Create navbar heading.
+        $name = 'theme_boost_union/navbarheading';
+        $title = get_string('navbarheading', 'theme_boost_union', null, true);
+        $setting = new admin_setting_heading($name, $title, null);
+        $tab->add($setting);
+
+        // Setting: Navbar color.
+        $name = 'theme_boost_union/navbarcolor';
+        $title = get_string('navbarcolorsetting', 'theme_boost_union', null, true);
+        $description = get_string('navbarcolorsetting', 'theme_boost_union', null, true);
```
lucaboesch, line 401: That should be navbarcolorsetting_desc, probably.

abias:
Well spotted, @lucaboesch !
I have indeed missed the `_desc` part in the string identifier!

abias:
Everything is solved and I have merged this now.

Amrita1991:
Hi @abias ,
I see changes related to this issue has been pushed to master branch but there has been no new release. We use a fork of the repo  which gets updated when there is a new release. We want to test this feature and decide on default colours for our university. Hence we were wondering are you planning for a new release soon? if yes, then when?

abias:
@Amrita1991 - thanks for asking.

Our release schedule is currently like this: Every weekend, if there were enough features / changes integrated since the last release, we backport the patches from master into the stable branches, tag a new release in Github and release the new version on moodle.org.

Thus, please expect the next release to be published by the beginning of next week.

Issue #228, opened 2023-02-12
Improvement: Add missing Behat tests for the H5P CSS feature


abias:
In https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/210, the H5P CSS feature was integrated without proper Behat tests. The mod_hvp plugin is already installed with GHA, but the tests to download and add the content types and then to test if the custom CSS is applied were not written.

This issue is there to pick this up and to finish the feature's commented tests in `theme_boost_union_looksettings_h5p.feature`

andremenrath:
@abias I started work on this and should get a PR ready later today.

andremenrath:
I am waiting for an upstream issue in the mod_hvp package to be fixed, otherwise testing this functionality for the hvp plugin is not possible https://github.com/h5p/moodle-mod_hvp/issues/487

abias:
Hi @Menrath ,

I reviewed your PR #229 and will merge it soon.

However, as you write, the scenario for testing mod_hvp is still missing.

Against this background, I have removed this code from moodle-plugin-ci.yml:

```
      - name: Add optional plugin dependencies
        run: |
          moodle-plugin-ci add-plugin --branch stable h5p/moodle-mod_hvp
          git -C moodle-plugin-ci-plugins/moodle-mod_hvp submodule update --init
```

It has to be added again when the scenario for mod_hvp is finally implemented.

abias:
PS: I do not think that anything will happen on https://github.com/h5p/moodle-mod_hvp/issues/487 anytime soon. But if you could build a patch for mod_hvp and could contribute the fix to mod_hvp, you could get things moving.

abias:
PR #229 has been merged now. This issue will be kept open for the remaining scenario which tests mod_hvp.

PR #229, opened 2023-02-16
Add behat tests for the customisation of the appearance of H5P activities

Menrath:
This is a WIP and partly solves issue #228.

It seems like adding a proper behat test for mod_hvp style adjusments is not possible at the moment.


abias:
Hi @Menrath ,

many thanks for working on this PR!

It's been really a while since you submitted this patch and I am sorry that it remained unhandled for so long.
I see that the PR is not fully finished yet, but I think that it could / should now be integrated at least partially.

Against this pbackground, I have reviewed and rework it in some aspects:

*  I rebased the branch onto latest master (of course)
* I changed the H5P CSS test to
```
.h5p-accordion {
        color: #FF0000 !important;
    }
``` 
as the purpose of this CSS is just to test if the CSS is applied and not to visually prettify the H5P content.
* I dropped the rem_css_value_to_px() function as it is not necessary anymore after the CSS code has been changed.
* I changed the `the element ".h5p-accordion" should have a "border-top-left-radius" of ".5rem"` step to `DOM element "#h5p-panel-content-0-0" should have computed style "color" "rgb(255, 0, 0)"` as this step is already available in Boost Union and does more or less the same - but I have gratefully adopted your non-jQuery version for checking the computed style to Boost Union's step definitions.
* I added steps for the "H5P content bank max width" scenario. This is not really related to your PR, but it is related to H5P as a whole and was missing as well.
* I removed the installation of mod_hvp from moodle-plugin-ci.yml until this plugin is supported as well.

I have pushed an updated branch to your repo and would then merge this PR as soon as the tests are green

Menrath:
@abias Thanks to you for your detailed review and picking this up!

Issue #232, opened 2023-03-05
Bug: Improve card header and footer of advertisement tiles with background images


abias:
If you do not use background images in advertisement tiles, the card header and footer gets a light grey background from Bootstrap: 
<img width="930" alt="grafik" src="https://user-images.githubusercontent.com/1092118/222955025-c635c574-cded-49a3-b424-66a905f42e37.png">

However, if you add a background image to the tiles, the card header gets unreadable. Additionally, you see a strange 1px border around the whole background image.
<img width="930" alt="grafik" src="https://user-images.githubusercontent.com/1092118/222954941-402cc59a-8470-4b3a-afb8-4404184c3b6d.png">

This should be improved like this:
<img width="929" alt="grafik" src="https://user-images.githubusercontent.com/1092118/222956902-9a1f8842-56c7-4f51-8a79-4750f91acb9a.png">

PS: Likewise, the card body also gets unreadable as soon as you add a background image. 
However, fixing this issue would be a greater piece of work in Boost Union as you can add arbitrary content to the cardf body. Thus, this should better be solved by the designer who is setting up the advertisement tiles.

lucaboesch:
Closed by #233

PR #233, opened 2023-03-05
Bugfix: Improve card header and footer of advertisement tiles with background images, solves #232

abias:
This PR realizes the look which has been posted in #232


lucaboesch (dismissed):
Looking good to me!
I think that's an unrisky thing to do.
Please merge.

Best,
Luca

lucaboesch (approved):

Issue #236, opened 2023-03-06
Navbar colour settings light font hover issue


james-pearce-bath-ac:
When the navbar colour is set to 'Primary color navbar with light font color' any dropdown menu items that are added via the 'Custom menu items' in the 'Theme settings' the text of the link when hovered stays as the light font colour rather than a contrasting dark colour, this would likely provide accessibility issues. Direct link hover states work as expected.

Navbar dropdown links before hover
![Brand-color-navbar-light-font-before-hover](https://user-images.githubusercontent.com/61325278/223100660-c8641750-6391-4ce9-af6f-eaca1a72ea03.png)

Navbar dropdown links when in hover state
![Brand-color-navbar-light-font-hover-issue](https://user-images.githubusercontent.com/61325278/223100541-4f12d64c-a089-4894-8489-078af1fc57c3.png)

abias:
Hi @james-pearce-bath-ac ,

many thanks for highlighting this issue.

I have to admit that I forgot the custom menu when I reviewed and extended the dark navbar PR.

I will fix this and submit a PR quite soon.

abias:
@james-pearce-bath-ac - I have created a PR on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/237.

If you have any chance, please give it a test-drive to see if I forgot any glitch which you have spotted.

abias:
#237 was merged now -> Closing.

Issue #235, opened 2023-03-06
Navbar colour setting - Edit mode indicator


james-pearce-bath-ac:
With the nagar colour options set to 'Primary color navbar with light font color' the edit mode indicator isn't very clear when set to on. The 'on' state uses the same primary colour as the colour to be used for the navbar background and so it isn't obvious that the edit mode is on.

**Example images below, our primary colour is a dark blue.**
Edit mode off
![edit-mode-off](https://user-images.githubusercontent.com/61325278/223095909-ef1a021c-00d6-4975-927d-c3a5fb17119e.png)
Edit mode on
![edit-mode-on](https://user-images.githubusercontent.com/61325278/223095914-75c14016-e669-4c62-9038-fb134aea1766.png)

Maybe adding a setting for the edit mode setting on background colour could resolve this issue and may be useful when not using Navbar colours as the setting could be used to specifically highlight the edit mode as a potential warning to users.

This issue isn't particularly critical but may offer some accessibility improvements.

abias:
Hi @james-pearce-bath-ac ,

many thanks for highlighting this issue.

In Moodle core, the edit switch uses the brand color as background color as soon as it is activated:
![grafik](https://user-images.githubusercontent.com/1092118/223180436-3c4cd86d-234c-46e0-8e4b-ca86d769809f.png)
![grafik](https://user-images.githubusercontent.com/1092118/223180474-a08a39f1-1e84-44d4-8b1b-76c194d584da.png)

We already made some small modifications to the edit switch within the dark navbar patch, but I have to admit that it was focused on the dark (grey) navbar and not the brand color navbar and it might have been too short-sighted.

I think for now we should just remove the brand color background from the edit switch on dark navbars:
![grafik](https://user-images.githubusercontent.com/1092118/223182063-289632a6-8ac6-4282-a82b-7eb0ebfe7ee2.png)
![grafik](https://user-images.githubusercontent.com/1092118/223182985-dd2411bc-8d4c-4008-9bca-670189950f5c.png)

![grafik](https://user-images.githubusercontent.com/1092118/223183146-4372f167-b23d-4436-9af9-934d70bc561b.png)
![grafik](https://user-images.githubusercontent.com/1092118/223183105-f88f9899-7da9-4048-ad62-fb7b8984bace.png)

![grafik](https://user-images.githubusercontent.com/1092118/223183352-b416aa2b-c8aa-4296-bd37-4622a6a221e4.png)
[...]

Issue #234, opened 2023-03-06
Navbar colour settings - light font colour makes the search close cross invisible


james-pearce-bath-ac:
The new navbar colour settings have added functionality that has been requested by several people in our institution and is very welcome thanks.

I have noticed when the navbar setting is set to a light font either with default dark background or primary colour there are a couple of issues with the search bar. 

1. The search close icon is picking up the css for icons to display using the light font colour option. This make the close icon impossible to view with those settings. This presents accessibility issues for use.   
![Search-bar-light-font issue](https://user-images.githubusercontent.com/61325278/223093899-862aa2ea-9292-441c-ac65-6583a6ca55ae.png)
2. Less critical the search bar background is white, see image attached. I would expect it to be using the background colour set in settings either the default colour or the primary colour. This might not be true of all users and the functionality of the search bar isn't affected so wouldn't be a priority compared to the above issue.

abias:
Hi @james-pearce-bath-ac ,

many thanks for highlighting this issue.

I have to admit that I forgot the search panel when I reviewed and extended the dark navbar PR.

I will fix this and submit a PR quite soon.

abias:
@james-pearce-bath-ac - I have created a PR on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/237.

If you have any chance, please give it a test-drive to see if I forgot any glitch which you have spotted.

abias:
#237 was merged now -> Closing.

PR #237, opened 2023-03-06
Solve multiple issues with dark navbar

abias:
This PR solves the issues with the dark navbar which were brought up in #234 #235 #236


abias:
As this is a more or less urgent bugfix, I will merge this myself without peer review.

Issue #245, opened 2023-03-17
Improvement: before_standard_html_head: ease child theming by checking for boost_union parentage


danowar2k:
before_standard_html_head currently checks if the currently selected theme is boost_union and doesn't do anything if it isn't.

To reuse the Font Awesome setting and the flavours a child theme currently has to at least copy the function and change it to its own name.

It would probably be child theming friendlier if boost_union checked if the current theme was boost_union or had boost_union as one of its parents. That way, the function wouldn't have to be reimplemented.

This checking could also be a setting ("ancestry allowed" or something), but that is for another time ;-)

I'll submit a PR for what I mean...

danowar2k:
A question while I am doing the PR: Why the condition 
```
$CFG->theme = 'boost_union'
```
and not
```
$PAGE->theme->name == 'boost_union'
```

Meaning, why is it important that the globally configured theme is boost_union and not just theme on this page?

abias:
This was solved in #246

PR #246, opened 2023-03-17
#245 - reduce code duplication for child theming

danowar2k:
Check for parentage in theme_boost_union_before_standard_html_head to prevent having to duplicate the function in child themes


abias (changes requested):
Hi @danowar2k ,

many thanks for submitting this improvement proposal!

I would love to integrate this improvement, but I am wondering what you are using the long path to load the whole theme config.

I would expect that `$THEME->parents` or `$PAGE->theme->parents` already contains the data which you want to check. I do not have a Boost Union child theme at hand to quickly test it myself. May I ask you to test it and adapt your patch if it works?

Thanks.

abias:
One more thing: I just spotted your question in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/245#issuecomment-1474115797 and would like to respond:
It's not important to use `$CFG->theme` at all.

I would be grateful if you could switch that to `$PAGE->theme->name` in your PR as well.

danowar2k:
I think I used theme_config::load() only because $CFG->theme was used and it could be possible that $PAGE->theme wasn't that globally configured theme.

$THEME on the other hand (I just checked) isn't a global. It was one once, though. Nowadays it's only set in the private function find_theme_config() (https://github.com/moodle/moodle/blob/880462a1685f710ed4f6628d31c2fd7e67e0df73/lib/outputlib.php#L2350-L2375).

I'll commit and rebase when I have time.

danowar2k:
Okay, I I've done the rebase and $PAGE->theme changes, hopefully correctly.

abias (approved):

abias:
Hi @danowar2k  - thanks for working again on this PR!
This looks fine to me now, merging...

PR #250, opened 2023-03-17
#249 - draft: don't force settings duplication, but allow it

### LINKED ISSUE #249: Improvement: Do not force child themes to reimplement any theme settings, but allow it
Right now, Boost Union implements a mix of forcing child themes to reimplement some settings and, in case of other settings, explicitly takes Boost Union's settings.

The first case could lead to unnecessary duplication of settings code.
The second case could maybe lead to duplication of function code if one would prefer to implement a theme's settings as "all settings for theme appearance should be configurable in a theme's settings pages).

I propose a function that returns a child theme's setting value if that setting exists in that child theme and otherwise returns Boost Union's setting value.

I'll prepare a pull request with that function (probably in lib.php or locallib.php, even if I like autoloaded classes wrapping this stuff, i.e. "class settingsutils"), and if the idea is accepted, I'd complete the PR with a bigger commit replacing all settings use with using this function.

danowar2k:
Added a method that allows child themes to duplicate boost union settings, but doesn't force them to do so, in case child themes want to just reuse boost union's configuration

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ moodle-theme_boost_union
 Changes
 -------
 
+* 2023-03-17 - Improvement: Do not force child themes to copy theme settings, but allow it.
+
 ### v4.1-r4
 
 * 2023-03-06 - Bugfix: Align the search bar with the dark navbar look, solves #234.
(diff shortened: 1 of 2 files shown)


abias:
Hi @danowar2k ,

many thanks for providing this contribution.
However, as discussed in chat, we will leave this PR pending for a while now as it would introduce some burden onto the theme maintainers and contributors in the long run.

Let's wait if your other PRs for child theme support turn out to be sufficient already.

Issue #260, opened 2023-03-20
Improvement: Prevent settings duplication in child themes by explicitly using theme_boost_union settings in theme_boost_union_get_pre_scss


danowar2k:
theme_boost_union_get_pre_scss can be called by either Boost Union itself or child themes of Boost Union.
The parameter $theme is thus not always Boost Union's theme_config, but can also be a child theme's theme_config.

As the method uses $theme->settings->SETTING_NAME , currently child themes have to duplicate Boost Union's settings used in there, because Moodle doesn't let child themes directly inherit parent theme's settings.

By changing all $theme->settings references to get_config('theme_boost_union') calls, Boost Union's settings are always used and child themes do not need to duplicate them.

abias:
This was solved in #261

PR #261, opened 2023-03-20
#260 - Explicitly use boost union's color settings

danowar2k:
Changed all instances of $theme->settings in theme_boost_union_get_pre_scss to get_config('theme_boost_union') to prevent child theme settings duplication

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -141,8 +141,8 @@ function theme_boost_union_get_pre_scss($theme) {
 
     // Prepend variables first.
     foreach ($configurable as $configkey => $targets) {
-        $value = isset($theme->settings->{$configkey}) ? $theme->settings->{$configkey} : null;
-        if (empty($value)) {
+        $value = get_config('theme_boost_union', $configkey);
+        if (!($value)) {
             continue;
         }
         array_map(function($target) use (&$scss, $value) {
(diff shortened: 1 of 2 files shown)


lib.php
```diff
@@ -141,8 +141,8 @@ function theme_boost_union_get_pre_scss($theme) {
 
     // Prepend variables first.
     foreach ($configurable as $configkey => $targets) {
-        $value = isset($theme->settings->{$configkey}) ? $theme->settings->{$configkey} : null;
-        if (empty($value)) {
+        $value = get_config('theme_boost_union', $configkey);
+        if (!($value)) {
```
abias, line 145: The goal of this PR is to replace `$theme->settings->{$configkey}` with `get_config('theme_boost_union', $configkey)`.
This is fine. However, as this part of the function is duplicated from upstream Boost Core and to avoid any hickups with edge cases, I would like to ask you to keep the rest of the login in these two lines unchanged unless you prove me that it must be changed.

lib.php
```diff
@@ -152,38 +152,38 @@ function theme_boost_union_get_pre_scss($theme) {
 
     // Overwrite Boost core SCSS variables which need units and thus couldn't be added to $configurable above.
     // Set variables which are influenced by the coursecontentmaxwidth setting.
-    if (isset($theme->settings->coursecontentmaxwidth)) {
-        $scss .= '$course-content-maxwidth: '.$theme->settings->coursecontentmaxwidth.";\n";
+    if (get_config('theme_boost_union', 'coursecontentmaxwidth')) {
```
abias, line 155: Is there a reason that you removed the `isset()` and `!empty` calls in this and the subsequent lines and rely on implicit PHP magic?

I would prefer to keep the isset() and `!empty` calls here just just to be sure.

abias (changes requested):
Hi @danowar2k ,

many thanks for working on this issue.
I have added two comments and would appreciate to get your feedback or code changes for them :)

Cheers,
Alex

lib.php
danowar2k: I'll change this like I did in #262 .

abias: Thank you :)

lib.php
abias: Touché!
Thank you very much for your elaboration, @danowar2k !

Against the background of your explanations, I am fine with your proposed changes.

abias (dismissed):

abias (approved):

Issue #262, opened 2023-03-20
Improvement: More consistency by using get_config in theme_boost_union_get_main_scss_content


danowar2k:
The config references of theme_boost_union_get_main_scss_content don't really need to be changed from $theme->settings to get_config() calls, because this function is only referenced by $THEME->scss. And that function is always called with the current theme and only by the theme_config of that theme.

Every child theme will probably (!) have its own main_scss function, so outside of a child theme explicitly calling the Boost Union main_scss function, there would be no need to reimplement the preset setting in child themes.

But still, to be consistent with other places switching to get_config() calls, this should be changed, too.

I'll open a PR.

wiebkemueller-hsh:
Hi @danowar2k can this issue be closed or are there remaining to dos that weren't implemented with the above mentionend merge?

abias:
This issue can now be closed as it is superseded by #267

PR #263, opened 2023-03-20
#262 - Use get_config in get_main_scss_content

danowar2k:
To be consistent across Boost Union, the final instances of $theme->settings are changed to get_config() calls.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -96,7 +96,8 @@ function theme_boost_union_get_main_scss_content($theme) {
     global $CFG;
 
     $scss = '';
-    $filename = !empty($theme->settings->preset) ? $theme->settings->preset : null;
+    $preset = get_config('theme_boost_union', 'preset');
+    $filename = !empty($preset) ? $preset : null;
     $fs = get_file_storage();
 
     $context = context_system::instance();
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2023-03-20 - Improvement: More consistency by using get_config in theme_boost_union_get_main_scss_content, solves #262.
+
 ### v4.1-r5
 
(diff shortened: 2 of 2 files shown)


lib.php
```diff
@@ -96,7 +96,7 @@ function theme_boost_union_get_main_scss_content($theme) {
     global $CFG;
 
     $scss = '';
-    $filename = !empty($theme->settings->preset) ? $theme->settings->preset : null;
+    $filename = get_config('theme_boost_union', 'preset');
```
abias, line 99: The goal of this PR is to replace `$theme->settings` with `get_config('theme_boost_union', ...)`.
This is fine. However, as this part of the function is duplicated from upstream Boost Core and to avoid any hickups with edge cases, I would like to ask you to keep the rest of the logic in this line unchanged unless you prove me that it must be changed.

abias (changes requested):
Hi @danowar2k ,

many thanks for working on this issue.
I have added a comment and would appreciate to get your feedback or code changes for it :)

Cheers,
Alex

lib.php
danowar2k: get_config($plugin,$value) returns either false, true or a string (which may be empty).

I removed !empty and the rest mainly because it is not necessary to set $filename explicitly to null, because `if ($filename)` returns false when $filename is null, false or an empty string, which is the desired behaviour.

But you're right, it would be easier when updating the upstream code to only replace that part.

By the way, looking at the function again, couldn't you, too, just reuse Boost's settings by doing this:

```
function theme_boost_union_get_main_scss_content($theme) {
    global $CFG;

    $scss = file_get_contents($CFG->dirroot . '/theme/boost_union/scss/boost_union/pre.scss');
    $scss .= theme_boost_get_main_scss_content(theme_config::load('theme_boost'));
    $scss .= file_get_contents($CFG->dirroot . '/theme/boost_union/scss/boost_union/post.scss');
    return $scss;
}
```

tl;dr
I'll change the code to leave the stuff in.
But there could be another way...?

abias:
Closing in favour of #268

Issue #267, opened 2023-03-23
Suggestion: Avoid duplicating Boost settings 'preset' and 'presetfiles'


danowar2k:
Boost Union duplicates Boost's 'preset' setting, without adding anything to it. You could avoid duplicating it entirely with appropriate changes to the code. I'll provide a PR.

wiebkemueller-hsh:
Hi @danowar2k can we close this  issue (superseeding #262) as well? I'm just trying to sort/clean up issues...in prep of the going live of the board...

abias:
Solved by #268

PR #268, opened 2023-03-23
#267 - Removed duplicated preset and presetfiles settings

danowar2k:
- Removed settings, preset files, and the language strings
- The main SCSS function just reuses boost's function so that boost's preset and presetfiles settings are used


danowar2k:
This lets users configure the preset and presetfiles (if they want) in Boost itself and removes duplicated code.

An extension of this would be to leave the 'preset' tab (general settings), but put a note and a link in there pointing to Boost's preset configuration section. Or maybe one just literally reuse and reprint Boost's preset settings? Although I don't know if it can be guaranteed that Boost's settings are always present before Boost Union's settings are setup...

danowar2k:
I think I got everything now. The commit now deletes the background image text placed unter "General Settings" in the README.md , too.

danowar2k:
> An extension of this would be to leave the 'preset' tab (general settings), but put a note and a link in there pointing to Boost's preset configuration section. Or maybe one just literally reuse and reprint Boost's preset settings? Although I don't know if it can be guaranteed that Boost's settings are always present before Boost Union's settings are setup...

I remember why I marked this as a draft. It's because of this comment. It would be something like:

* Keep the General settings/presets tab
* Add a desciption telling "Please set preset and presetfiles in the Boost config linked here

danowar2k:
I'm looking through my (ages old) contributions. This PR seems to have been finished by me, but is still open and no review was done, it seems. Am I missing anything or did we talk about this somewhere else? I'm unsure how to proceed...

danowar2k:
Okay, I've cleaned up some stuff and changed some filenames. 

It looks to me there are now a few things to do:
- A general review
- Decide whether the settings should stay this way (General page with preset and presetfiles gone) or whether I should re-add it with a link to the Boost settings

Am I missing something?

abias:
Hi @danowar2k ,

many thanks for this PR and sorry that it remained unhandled for so long.

I have now processed it and agree that the preset setting could / should be removed from Boost Union. I have especially seen that it seems to be broken currently in Boost Union which makes even more sense to remove it from Boost Union and just use the one from Boost Core.

I have now rebased your branch and have added a review change commit which keeps the "general settings" tab and a stub description which points to the theme presets in Boost Core - as this is was you asked for and what makes sense to me.

I also added comments to theme_boost_union_get_main_scss_content instead of renaming pre.scss and post.scss. This was done because I simply do not want to break all the PRs which are currently open and forks which may exist out there and which modified post.scss.

As an addition,  I removed the preset settings in the database within an upgrade step and I added a Behat test to cover the inherited preset setting in Boost Union.

As soon as the tests are green, I will merge this PR into the childtheme branch which acts as an integration buffer for the next few days as we are working on multiple child theme support improvements which should land in master all at once.

abias:
The tests are not green, but only because of #559 which is unrelated to this PR.
I will merge the PR then

abias:
The PR was squashed and merged locally and pushed into the childtheme branch.

PR #280, opened 2023-04-13
Fix js error when there is no edit toggle

t-schroeder:
On pages where there is no edit toggle (most pages for students) you get a JavaScript error:

> Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')

This simply adds a check for whether an edit toggle is even there.


t-schroeder:
Hi @lucaboesch 

you're right. I didn't see that other issue or pull request. Closing this one.

Issue #273, opened 2023-04-05
Dark navbar hides notification menu buttons


t-schroeder:
If you use the dark navbar setting the buttons in the notification menu appear white on white background:

![image](https://user-images.githubusercontent.com/24255259/230100979-f75180d8-46ba-4de4-9154-620a385649ab.png)

This is what they should look like:

![image](https://user-images.githubusercontent.com/24255259/230101239-73f18367-1915-4ad2-bb20-66815a32471c.png)

The issue comes from this CSS rule: https://github.com/moodle/moodle/blob/v4.1.2/theme/boost/style/moodle.css#L12377-L12379

So I'm wondering: Should this even be fixed in this theme or rather in Moodle itself?

alina-kiz:
I have the same problem

abias:
This has been solved by #303

PR #303, opened 2023-05-17
Improvement: Improve SCSS for dark navbar and primary color navbar, solves #273.

wa-bea:
This pull request solves issue #273.
Furthermore, it improves the look of the primary color navbar with the dark text, changes the color of the language menu for not logged-in users. 

For clarification, what this PR is about: 

Before
image

After
image

Before
image

After
image

Before
image

After
image

Before
image

After
image

Before
image

After
image

Before
image

After
image

Before
image

After
image

### DIFF
--- a/scss/boost_union/post.scss
+++ b/scss/boost_union/post.scss
@@ -112,6 +112,20 @@
     .login a:hover {
         color: $white;
     }
+
+    /* Change the color of the notification action buttons to dark grey.
+       This has to use !important as Moodle core already uses !important for the icon colors. */
+    .popover-region-container a .icon {
+        color: #343a40 !important; /* stylelint-disable-line declaration-no-important */
+    }
+
+    /* Change the color of the language selector link in the navbar to a light grey and to white when hovered. */
+    .langmenu .dropdown .dropdown-toggle {
+        color: #c8c8c8;
+        &:hover {
+            color: $white;
+        }
+    }
 }
 
 /* Tweak the dark navbar with light font color a little bit more. */
@@ -132,6 +146,17 @@
     /* Change the background color of the search panel in the navbar. */
     #searchform-navbar {
         background-color: #343a40;
+    }
+
+    /* Change the color of the language selector drop down items when hovered, focused and active to dark grey
+       as this would be presented in the primary color otherwise. */
+    .langmenu {
+        .dropdown-item:active,
+        .dropdown-item:hover,
+        .dropdown-item:focus,
+        .dropdown-item:focus-within {
+            background-color: $dark;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


scss/boost_union/post.scss
```diff
@@ -154,14 +179,26 @@
...
         background-color: $primary;
     }
+
+    /* Change the color of the active nav items's bottom border to white as this would be presented in the
+   primary color otherwise. */
+    .nav-link.active {
+        border-bottom-color: $white;
+    }
+
+    /* Change the color of the language selector icon to a light grey as well.
```
lucaboesch, line 189: Try to prepend this by

`/*csslint important:false*/`

I think you give the rationale to use this in the comment. The above line would then shut csslint quiet.
Here's a usage example: https://github.com/moodle/moodle/blob/MOODLE_402_STABLE/lib/yuilib/3.18.1/cssnormalize/cssnormalize.css#L565

Thanks for your contribution, @wa-bea !

Best,
Luca

scss/boost_union/post.scss
```diff
@@ -112,6 +112,20 @@
     .login a:hover {
         color: $white;
     }
+
```
lucaboesch, line 115: Try to prepend this by

/*csslint important:false*/

I think you give the rationale to use this in the comment. The above line would then shut csslint quiet.
Here's a usage example: https://github.com/moodle/moodle/blob/MOODLE_402_STABLE/lib/yuilib/3.18.1/cssnormalize/cssnormalize.css#L565

Thanks for your contribution, @wa-bea !

Best,
Luca

scss/boost_union/post.scss
```diff
@@ -194,6 +231,21 @@
...
+    /* Change the color of the login link in the navbar to dark grey and to darker grey when hovered. */
+    .login,
+    .login a {
+        color: rgba($black, .6);
+        &:hover {
+            color: rgba($black, .9);
+        }
+    }
+
```
lucaboesch, line 243: Try to prepend this by

/*csslint important:false*/

I think you give the rationale to use this in the comment. The above line would then shut csslint quiet.
Here's a usage example: https://github.com/moodle/moodle/blob/MOODLE_402_STABLE/lib/yuilib/3.18.1/cssnormalize/cssnormalize.css#L565

Thanks for your contribution, @wa-bea !

Best,
Luca

lucaboesch (changes requested):

scss/boost_union/post.scss
abias: Hi Luca,

thank you for this hint how to fix the linter warning!

However, may I propose to use a slightly different syntax which we already use in Boost Union?
If you look at https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/79e568dd9c5a3c81440edd836c51755f2c9b2344/scss/boost_union/post.scss#L77, you will see how it's done and Beata can directly adopt it to her code.

scss/boost_union/post.scss
abias: Please see my comment above :)

scss/boost_union/post.scss
abias: Please see my comment above :)

scss/boost_union/post.scss
lucaboesch: Ah, yes, the inline one!
That's much more elegant, correct.

lucaboesch (approved):
That looks good to me!
Thanks @wa-bea

Issue #305, opened 2023-06-02
Moodle phpunit tests fail with boost_union installed on 4.1 due to customfiletypes


radzisze:
Moodle 4.1.3
Boost union v4.1-r7

When running Moodle's phpunit tests with boost_union installed I noticed this Moodle core test started failing: 
```
core\filetypes_test::test_cleanup
Failed asserting that object of class "stdClass" does not have attribute "customfiletypes".
/home/travis/build/isteclass/eclass/lib/tests/filetypes_test.php:217
/home/travis/build/isteclass/eclass/lib/phpunit/classes/advanced_testcase.php:80
phpvfscomposer:///home/travis/build/isteclass/eclass/vendor/phpunit/phpunit/phpunit:97
```
It appears that Moodle assumes that customfiletypes should be empty at the start of that test, and customfiletypes are added in function theme_boost_union_register_webfonts_filetypes() of locallib.php of boost_union. There is a nice check at the start of this function already to see if customfiletypes are set in config.php that just returns false. Perhaps adding a check for if phpunit is running or something could be added there (" || PHPUNIT_TEST" maybe?) or in that function to prevent adding customfiletypes during phpunit test runs as well?

lucaboesch:
Dear @radzisze I crafted https://github.com/lucaboesch/moodle-theme_boost_union/tree/phpunitcustomfiletypes based on your proposal. Could you check out if that works for you?

abias:
This was solved by #306

PR #306, opened 2023-06-03
Fix PHPUnit runs with Boost Union, solves #305

lucaboesch:
Thanks @radzisze for the report and the proposal on how to tackle this!

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -786,7 +786,7 @@ function theme_boost_union_get_webfonts_extensions() {
 function theme_boost_union_register_webfonts_filetypes() {
     global $CFG;
 
-    // If customfiletypes are set in config.php, we can't do anything.
+    // If customfiletypes are set in config.php or PHP tests are running, we can't do anything.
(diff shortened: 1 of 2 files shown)


locallib.php
```diff
@@ -787,7 +787,7 @@ function theme_boost_union_register_webfonts_filetypes() {
     global $CFG;
 
     // If customfiletypes are set in config.php, we can't do anything.
```
abias, line 789:
```suggestion
    // If customfiletypes are set in config.php or PHP tests are running, we can't do anything.
```

abias (changes requested):
Hi @lucaboesch ,

many thanks for working on this issue!
I think the solution is spot-on and it's great that tests are passing again for @radzisze .

May I just ask you to extend the comment which explains the if-clause?
Afterwards, you can cirectly merge the PR :)

Cheers,
Alex

abias:
Hi @lucaboesch ,

I have now added this small change myself, have rebased the branch onto latest master and will merge the PR now.

Cheers,
Alex

Issue #269, opened 2023-03-27
Wrong footer and footnote placement caused by new block region feature


larsbonczek:
There is a regression caused by the new block region feature introduced in #118. The new `.main-inner-wrapper` element added to [drawers.mustache](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/6a100b414e5647ec46f68033ba78fec3524a6010/templates/drawers.mustache#L145) breaks at least one CSS rule:

[/theme/boost/scss/moodle/layout.scss#L48](https://github.com/moodle/moodle/blob/09fd61415e45f2f1410d75cce093b852419dcf6e/theme/boost/scss/moodle/layout.scss#L48)

The `flex` property needs a flex parent. Its previous parent, the `#page` element, is a flex container. The new `.main-inner-wrapper` element is not. This means that the `flex-grow: 1` property is not applied to the `.main-inner` element. This leads to a wrong placement of the footer and footnote elements on short pages as seen in this example:

![Screenshot from 2023-03-27 16-53-54](https://user-images.githubusercontent.com/44197504/227977978-d32fc586-89b1-4f8c-8b05-10a75ca39e33.png)

abias:
Many thanks for reporting this regression, @larsbonczek !
This was fixed in #309 now.

PR #309, opened 2023-06-14
Bugfix: Fix footer and footnote placement, solves #269.

wa-bea:
Adding the flex-grow-1 class to the new .main-inner-wrapper element fixes the problem.

Before
image

After
image


abias (approved):
Many thanks for working on this issue, Beata!
The fix is spot-on from my point of view and will be merged now.

Issue #307, opened 2023-06-13
Feature: Mark External links 


DHBW-Katja:
Similar to Issue #164: 
It would be nice to mark all links in Moodle that link to other sites (external from our domain) with an icon, i.e. box with arrow up. 

Just as an idea for CSS: 
a[href^="http:"]::after,
a[href^="https:"]::after {
font-family: fontawesome;
content: "\f08e";
font-size: 90%;
padding-left: 3px;
}
a[href*="dhbw.de/"]::after {
content: "" !important;
padding-left:  0 !important
}

abias:
Many thanks for submitting this proposal, @DHBW-Katja !
This was solved in #310 now.

PR #310, opened 2023-06-14
Feature: Add admin option to mark external links, solves #307.

abias:


lucaboesch (commented):
I have the impression that this change can not do other than completely break any RTL language compatibility (if there were one).
Has this been tested?
Is there the possibility to make those changes work for RTL languages as well?

abias:
Hi @lucaboesch ,

many thanks for this helpful insight! We completely missed RTL during this development unfortunately.
I have added a change which adds separate CSS rules for RTL and for LTR which solved this issue in my tests.

Additionally, I have seen that Moodle adds a hardcoded external-link icon to the "services and support" link in the questionmark menu. This becomes obsolete with this feature. We remove it with the sledgehammer CSS rule as well now.

ksteitz:
looks like there is a second "hardcoded external-link icon" in the questionmark menu: "Help and documentation"

grafik

Any idea why the icon isn´t workin in our Moodle 4.2+ (Build: 20230512) ?
grafik

thanks and cheers!
Klaus

Issue #321, opened 2023-06-20
Improvement: Shrink description_format column size in theme_boost_union_flavours table


abias:
The description_format column in the theme_boost_union_flavours table has a size of 10 (see https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/db/install.xml#L12), but in fact it only needs to have size of 2 given the data which is stored there.

It would be worthwhile to shrink this column in db/install.xml and db/upgrade.php

Harshal662:
I can do that

abias:
@Harshal662 - that would be awesome!
Please don't hesitate to create a PR for this issue.

lucaboesch:
Fixed by #505.

PR #324, opened 2023-06-21
length change

Harshal662:
changed length from 10 -> 2

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -9,7 +9,7 @@
         <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
         <FIELD NAME="title" TYPE="char" LENGTH="255" NOTNULL="true" SEQUENCE="false"/>
         <FIELD NAME="description" TYPE="text" NOTNULL="false" SEQUENCE="false"/>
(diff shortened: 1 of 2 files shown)


abias (changes requested):
Hi @Harshal662 ,

many thanks for submitting a fix for issue #321

However, I am afraid that your solution is only partly right.

The change in db/install.xml is fine, but the change in db/upgrade.php is not correct.
You must not modify existing upgrade steps as they won't be executed if the plugin is already installed in this version. You have to raise the plugin version in version.php and have to write a step in db/upgrade.php which changes the length of the column for this version upgrade. The Moodle XMLDB editor (see https://docs.moodle.org/dev/XMLDB_editor) will help you to edit the table and to create the code for upgrade.php.

Looking forward for the next iteration of your patch!

Cheers,
Alex

abias:
Hi @Harshal662 - Do you see any chance that you can do the changes which I asked your above?
Otherwise, I will have to close the PR and someelse can tackle the issue.

abias:
Hi @Harshal662 ,

I am sorry to close your PR as there wasn't any update from you anymore.
I have submitted a proper PR on #505 myself.

Cheers,
Alex

Issue #323, opened 2023-06-21
Bug: "theme_boost_union | markexternallinks" double labeling in the footer


stdoma:
Hello!
I have activated the setting to mark external links.
The help link is already marked in the footer by the system.
As a result, the link is marked twice.

(I've customized the footer, so it looks a little different than the default.)
![Screenshot 2023-06-21 153826](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/92015860/9070be71-1292-45c2-9ee0-9411ef104b6e)

ralf-krause:
Yes, the external links are double labeled. In my case I get no icon for the label from Boost union.

<img width="362" alt="moodle-copyright" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/27956839/3eeddb9a-8227-4849-a7d0-7fad3fe51148">

<img width="777" alt="moodle-server-environment" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/27956839/7a116dc5-3321-4389-8654-760adf350767">

ksteitz:
This behaviour is a bug affecting 4.2 only, see also https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/310#issuecomment-1602791784
It is caused by the change to font awesome 6. 
There will be a fix soon...

ralf-krause:
Okay, we get the empty icons only with Moodle 4.2. 
But you marked this complete doublelabeling issue for 4.2 only. This is incorrect. My screenshot shows Moodle 4.0.8. The external links are doublelabeled.

<img width="601" alt="moodle-server-environment-408" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/27956839/e6653306-ef6f-447c-887b-ca86935b734f">

ksteitz:
The "Bug: Font Awesome external link icon broken in Moodle 4.2+" is now in a separate issue: 
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/327

abias:
Thank you all for raising and discussing this issue.

I have just pushed 48b3145497fb3ed9f91b738fadb01b99c31d99d5 for Moodle 4.2 and 5d14ac5d66209240bb1633c402c83cefce670182 for Moodle 4.1 which should hopefully fix this issue.

Cheers,
Alex

Issue #327, opened 2023-06-26
Bug: Font Awesome external link icon broken in Moodle 4.2+


ShUl0w:
Due to a major upgrade of Font Awesome from 4.x to 6.x in Moodle 4.2, some Font Awesome icons are not properly displayed:
![image](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/37832993/80831569-1191-4954-bae8-cac163e05868)

I have already begun work on a bugfix, which should be ready shortly. However, there is one obstacle to discuss: from Font Awesome 6 onward, the "normal" font-style as well as the corresponding font-weight "400" appear to be a pro feature, meaning they are absent from Moodle, at least from what I can tell. Font Awesome 6 only seems to offer the font-weight "900" style.

Thus, a fallback for Moodle <= 4.1 needs to be implemented when moving up to Font Awesome 6. This brings with it some questions regarding the cleanliness of the codebase when moving forward.

I therefore propose a "quickfix" to solve the missing icon issue asap, and then figure out how to move forward. For the time being, we need to decide if a bold icon for Moodle <= 4.1 is okay, or if we should implement version checks to retain the "original" look (see screenshot below for comparison; left 4.2, right 4.1).

![image](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/37832993/6c4b5473-5c03-49d7-a3e7-2c0c50b4a15c)

ShUl0w:
I've opened a draft PR, see #328

ShUl0w:
As an idea, we could also implement a kind of boost-union font-face, which should - unless my CSS-knowledge eludes me - provide even cleaner fallback capabilities:

`
@font-face {
  font-family: "Boost Union Font Awesome";
  src: local("Font Awesome 6 Free"), local("FontAwesome");
  font-weight: 900;
}
`
Since I'm not as experienced in Moodle development yet, is it maybe possible to pass a variable to the SCSS compiler with which we could set the font-weight dynamically, depending on the Moodle version?

abias:
Thank you all for raising and discussing this issue.

I have just pushed 48b3145497fb3ed9f91b738fadb01b99c31d99d5 for Moodle 4.2 and 5d14ac5d66209240bb1633c402c83cefce670182 for Moodle 4.1 which should hopefully fix this issue.

Cheers,
Alex

PR #274, opened 2023-04-06
 #163 #164 - Added settings to mark broken links and mailto links

matthias-buttgereit:
- Added two settings in Look -> SCSS
- Added language strings
- Added functions to locallib.php
- Calling those functions in lib.php

PR #328, opened 2023-06-26
Bugfix: Moodle 4.2 / FontAwesome 6 broken external link icon, solves #327

ShUl0w:
I've implemented a bugfix for #327, which currently offers backwards-compatability but causes the external link icon to be a bit bolder (900 vs 400) in Moodle versions <= 4.1. This can be handled differently, e.g. by implementing a version check and still needs further discussion in #327.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2023-06-26 - Bugfix: Fixed broken FontAwesome 6 external link icon in Moodle 4.2+, solves #327.
+
 ### v4.1-r8
 
 * 2023-06-18 - Improvement: Split "Branding" tab into "Site branding" and "Activity branding", solves #315.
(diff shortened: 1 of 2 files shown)


ShUl0w:
As of right now, this fix would only solve #327 and a part of #323, however, #274 would still break in 4.2. I've proposed another way of handling this in #327, which probably requires further discussion

abias:
Hi @ShUl0w ,

many thanks for working on a fix for the given issue!

I appreciate your work, but think that this PR can't / should not be merged as it is now as Boost Union is maintaint in separate branches for each Moodle major version. Thus, we don't need backwards-compatibility within a branch.

I am currently finishing the 4.2 release of Boost Union and, doing this, will also fix #323 and #327 for 4.2. Afterwards, I will backport the fix to the 4.1 branch.

I will comment in these issues as soon as the solution is pushed to Github.

Thanks again,
Alex

abias:
PS: @ShUl0w - Please let me clarify that the fix which I will push soon will basically pick up your fully valid and working bugfix, but will just be integrated into the 4.2 and 4.1 branches separately. Thanks again for your work :)

Issue #325, opened 2023-06-22
Bug: $blockregionoutsiderightwidth and $blockregionoutsideleftwidth are missing default values in scss


danowar2k:
If somehow the values for blockregionoutsiderightwidth or blockregionoutsideleftwidth are not set from within the GUI or the values are set as empty, the theme can't be built because the variables don't have default values. (see https://sass-lang.com/documentation/variables#default-values)

These default values should probably be set at the beginning of post.scss or in pre.scss, depending on where the scss generated by setting the variables is placed.

danowar2k:
Same as #326.

abias:
Thank you for raising this issue!
I cannot imagine a case how the prerequisites for this bug could be triggered, but there was a gap in the code which sets the SCSS variables which I fixed in PR #329.

abias:
This was fixed in #329

PR #329, opened 2023-06-26
Bugfix: If block settings were not set, the SCSS was not compiled, solves #325.

abias:
This PR should solve #325 and #326. There is no obvious case which would trigger the prerequisites for these issues up to now, but if these prerequisites are there, the resulting bug is avoided.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2023-06-26 - Bugfix: If the settings blockregionoutsiderightwidth or blockregionoutsiderightwidth were not set for any reason, the SCSS was not compiled, solves #325.
+
 ### v4.1-r8
 
 * 2023-06-18 - Improvement: Split "Branding" tab into "Site branding" and "Activity branding", solves #315.
--- a/lib.php
+++ b/lib.php
@@ -200,14 +200,23 @@ function theme_boost_union_get_pre_scss($theme) {
         $scss .= $activityiconscss."\n";
     }
 
-    // Set custom Boost Union SCSS variables.
-    if (get_config('theme_boost_union', 'blockregionoutsideleftwidth')) {
-        $scss .= '$blockregionoutsideleftwidth: '.get_config('theme_boost_union', 'blockregionoutsideleftwidth').";\n";
+    // Set custom Boost Union SCSS variable: The block region outside left width.
+    $blockregionoutsideleftwidth = get_config('theme_boost_union', 'blockregionoutsideleftwidth');
+    // If the setting is not set.
+    if (!$blockregionoutsideleftwidth) {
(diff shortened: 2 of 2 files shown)


danowar2k:
Suggestion: Always write $blockregionoutsideleftwidth, put the value in another variable with either get_config or the default as a value and append that. Less if/else... I'll follow up with concrete code.

danowar2k:
```
$blockregionoutsideleftwidthvalue = get_config('theme_boost_union', 'blockregionoutsideleftwidth') ?? '300px';
$scss .= '$blockregionoutsideleftwidth: '.$blockregionoutsideleftwidthvalue.'\n';
```

Or something like this. I hope I remember the ?? operator correctly.

danowar2k:
Sorry, ?? would need the value to be null, but if get_config doesn't find a value it returns null. So, slightly longer:

```
$blockregionoutsideleftwidthvalue = get_config('theme_boost_union', 'blockregionoutsideleftwidth') ? get_config('theme_boost_union', 'blockregionoutsideleftwidth') : '300px';
$scss .= '$blockregionoutsideleftwidth: '.$blockregionoutsideleftwidthvalue.'\n';
```

abias:
@danowar2k 

> Suggestion: Always write $blockregionoutsideleftwidth, put the value in another variable with either get_config or the default as a value and append that. Less if/else... I'll follow up with concrete code.

I understand what you are proposing, but I tend to stick with my code because it is more readable / verbose at first sight. Or would your code have any other advantages in terms of performance, correctness or covering additional edge cases which I am missing?

> Of course, since now the 300px default is used both in lib.php and settings.php, maybe that default value should become an additional constant referenced by both...

I am aware of the fact that these 300px are now hardcoded in more than one place. 
However, the 300px value in lib.php is really only used if something has already gone wrong in the theme and it is there to avoid that even more goes wrong. Personally, I would prefer keeping both values hardcoded separately instead of adding another constant to lib.php which is loaded into memory on every page load uselessly.

What do you think?

danowar2k:
No advantages in those terms, just more DRY and less lines to manage.

To be honest, I personally would prefer an autoloaded class UnionConstants over all the defined constants in lib.php. I don't know if that would have any advantages over the current way, though...

danowar2k:
For what it's worth, my suggestion mainly intended to remove the duplicated string concatenation. The ? : I just used because I like the conciseness of it.

abias:
> No advantages in those terms, just more DRY and less lines to manage.

Point taken.

> To be honest, I personally would prefer an autoloaded class UnionConstants over all the defined constants in lib.php. I don't know if that would have any advantages over the current way, though...

Well, the underlying architecture can always be improved, but we have to work with the tools we have now.

> For what it's worth, my suggestion mainly intended to remove the duplicated string concatenation. The ? : I just used because I like the conciseness of it.
> 
> Variation:

Allright, I agree with your last proposal.
I will update the PR and will then - as I understand your feedback as a peer review - merge the PR.

Cheers,
Alex

Issue #330, opened 2023-06-30
Bug: In master, foreach() argument must be of type array|object warning appears. 


lucaboesch:
This appears in (PHP8.1) master (hence Moodle 4.3pre):

`Warning: foreach() argument must be of type array|object, bool given in …/theme/boost_union/locallib.php on line 1203`

when a site is rebased to a weekly release and admin/index.php is called and the site wants to update.

<img width="1512" alt="boost_union_warning" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/377279/458e1423-f0ac-4634-843a-ed451da949bf">

lucaboesch:
I think I was on a old branch. Didn't happen when updating Moodle 4.2.

lucaboesch:
No, it is legit. But it appears only if no CSS cached folder is present.

abias:
This was fixed in #331

PR #331, opened 2023-06-30
Bugfix: Warning if no CSS cached folder is present on updating, solves #330.

lucaboesch:
I do not know why my system did not have a CSS cached folder though…

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


abias:
Thank you, @lucaboesch , for fixing this debug notice.
I agree that it's strange that this notice is shown at all, but your fix is fine for me.

I just rebase the branch to latest master and added a comment over your new if-clause.

PR #340, opened 2023-07-17
Scrollspy loading fix

mwehr:
This adds loading of scrollspy only on course pages.


abias:
Hi @mwehr  ,

many thanks for working on this PR!

It's been really a while since you submitted this patch and I am sorry that it remained unhandled for so long.

I have now finally processed it and have as a first step squashed and rebased the branch onto latest master now.

Regarding your solution approach, I agree that checking for the existence of `.section.main` (i.e. to check if any course section exists on the current page) is helpful, but in fact the AMD module should not be added to non-course pages at all. But currently it is added to all pages and, in your patch for layout/includes/scrollspy.php, you are just excluding the frontpage.

That's why I have made two additions to layout/includes/scrollspy.php:
1. I check for the page layout as well and add the AMD module only if the pagelayout is `course`
2. I changed your check if we are on the frontpage from the hardcoded ID 1 to `SITEID` (which is a constant in Moodle core which carries the frontpage course ID - which could in fact be unequal to 1). However, this check should never be checked now as the frontpage does not have the pagelayout `course`.

I have pushed an updated branch to your repo and would merge the PR as soon as the tests are green.

abias (approved):

PR #341, opened 2023-07-17
Adds remembering of active admin tab on page reload.

### LINKED ISSUE #468: Remembering of active boost union admin sub-tab after save and page reload
If you change a value in the sub-tabs of the Boost Union Admin setting, the first tab is always selected after reloading the page and not the current tab in which the change was made.

It would be helpful if the current Boost Union sub-tab is opened again after saving and reloading the admin settings page.

mwehr:
This adds remembering the active tab in boost union admin settings on page reload. resolves #468


abias:
Hi @mwehr ,

many thanks for submitting the PR and sorry that it remained unhandled for so long.
We are just ramping up or PR reviewing again after we have established some more processes to keep the overview over our work.

As a result, we have created a board on https://github.com/orgs/moodle-an-hochschulen/projects/4/views/1 and have established a policy that every PR needs a corresponding issue which it solves.

May I ask you to quickly create a Github issue with a brief explanation of the goal which this PR tries to solve? I will then be able to add your issue to the board and review your PR.

Thanks,
Alex

abias (dismissed):
Hi @mwehr ,

many thanks for this contribution and sorry that it remained unhandled for so long!

This is a really helpful addition to Boost Union for all admins who have to handle the same settings page again and again as they are doing multiple theme configurations in a row or want to test multiple values of the same setting.

I just rebased your PR branch onto the latest master and have reviewed your PR.

The review was quite easy. I just made some tiny refactorings and made the comments more verbose to match the current Boost Union codebase.
I also added a version bump (as a JS module has been added and the JS cache should be cleared after installing this patch), an entry in CHANGES.md and a Behat scenario.

I did _not_ add an admin setting to enable this feature (like we normally do) and did not add a note to README about this feature as well. My understanding is that this is not necessary as it's a 100% admin-only feature and should "just work".

I will merge the PR as soon as the tests are green again.

Cheers,
Alex

abias (approved):

Issue #351, opened 2023-07-27
Improvement: Add 'aboutus', 'offers', 'page1', 'page2' and 'page3' static pages


wa-bea:
It would be nice if we would have additional prepared static pages which can be easily filled by admins:

- About us
- Offers
- Generic page 1
- Generic page 2
- Generic page 3

abias:
This was solved by #353

PR #353, opened 2023-08-02
Improvement: Add more static pages, solves #351.

wa-bea:
This patch adds five additional static pages:

- About us
- Offers
- Generic page 1
- Generic page 2
- Generic page 3

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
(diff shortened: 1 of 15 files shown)


abias:
Thank you very much @wa-bea for working on this PR!

Adding some more static pages makes sense. I agree with the addition of the about us and the offers page and the generic pages also make sense for installations who need even more pages without changing the Boost Union source code.

Your patch is spot-on.
I just rebased your PR onto the latest master and changed the order of the pages in the language pack, on the settings page and so on based on the order of the page links in the footnote and the footer.

As soon as the tests are green, I will merge the PR.

abias (approved):

abias (approved):

PR #354, opened 2023-08-03
Add js and css to use local_och5p with mod_hvp

elke-hsh:
local_och5p offers a button to select opencast videos for H5P.InteractiveVideo. 

The plugin local_och5p creates a renderes.php with needed css and js in the theme directory. But as theme_boost_union  already integrates a renderers.php with similar functions since version v4.1-r2 it does not work properly.

This pull request adds the css and js if mod_hvp and local_och5p as well as H5P.InteractiveVideo is installed.

I would like to see this in version 401 as well. Should I create a pull request for that branch, too?

### DIFF
--- a/renderers.php
+++ b/renderers.php
@@ -79,6 +79,7 @@ if (file_exists($CFG->dirroot.'/mod/hvp/renderer.php')) {
              * @param string $embedtype How the H5P is displayed.
              */
             public function hvp_alter_styles(&$styles, $libraries, $embedtype) {
+                global $CFG;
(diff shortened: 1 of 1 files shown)


abias:
Thank you, @elke-hsh , for submitting this PR!

We have to be a little bit careful with adding support for third party plugins in Boost Union, but I am aware that local_och5p simply needs a renderer modification in the used theme and it should be ok to add this to Boost Union directly.

However, your PR is superseded by #359 by @ferishili who extended your patch and added support for the core H5P module as well.

PR #400, opened 2023-09-25
Changed 'imprint' to 'legalnotice' across pages

VivekChatterjee:
Changed 'imprint' to 'legalnotice' across Various pages.


danowar2k:
I suggest adding an upgrade step to convert the imprint settings values to the legal notice settings values.

Issue #129, opened 2022-11-03
Feature: Improve usability of course management page


abias:
In https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus/issues/69, @stemiwe wrote for Boost Campus:

> We improved the usability of our course management page by 3 changes, I think it would be great to include them in boost_campus:
> 
> 1. adding the search courses box to the top of the management view
> 2. showing both short name and full name for a course in the course area
> 3. adding a direct link to the selected course to the course view area
> 
> ![managementview](https://user-images.githubusercontent.com/13886080/70710254-0e372080-1cdf-11ea-80d9-b466a2f039dd.png)
> 
> we accomplished this by changes in
> /course/management.php
> /course/classes/management_renderer.php
> /course/classes/management/helper.php
> /course/classes/list_element.php
> 
> in /course/management.php add this line before "echo $renderer->management_form_start();"
> 
> ```
> echo $renderer->course_search_form($search);
> ```
> 
> in /course/classes/management_renderer.php add a view button in the function search_listitem_actions:
> 
> ```
>        // View.
>        $actions[] = $this->output->action_icon(
>            new \moodle_url('/course/view.php', array('id' => $course->id)),
>            new \pix_icon('i/course', \get_string('view')),
>            null,
>            array('class' => 'action-view')
>        );
> ```
> 
> In /course/classes/management/helper.php add a view button in function get_course_listitem_actions:
> 
> ```
>        // View. 
>            $actions[] = array(
>            'url' => new \moodle_url('/course/view.php', array('id' => $course->id)),
>            'icon' => new \pix_icon('i/course', \get_string('view')),
>            'attributes' => array('class' => 'action-view')
>        );
> ```
> 
> In /course/classes/list_element.php change function get_formatted_name:
> 
> ```
> return format_string($this->__get('shortname'), true, $this->get_context()) . ' - ' . format_string($this->__get('fullname'), true, $this->get_context())
> ```

While the first item (the search box) is there now in Moodle 4.x, the second and the third proposals are still valid for Boost Union and it would be worthwhile to solve it.

abias:
The second goal can be achieved with the Moodle core setting $CFG->courselistshortnames.

I will work on implementing the remaining third goal now.

lucaboesch:
I think this change 'adding a direct link to the selected course to the course view area' would be great to include them in Moodle Core.
Please add an issue to https://tracker.moodle.org/, @stemiwe.
[...]

PR #417, opened 2023-10-09
Improvement: Add a direct 'view course' icon on the course management pages, solves #129.

abias:


lucaboesch (changes requested):
Oh my, I'm sorry to rain on the parade. It's probably also due to time we let pass to develop this. But I just noticed we call that 'thing' all sort of names: "link", "icon", "button".
That's no good.
We should stick to something and then call it the same way throughout everything (also renaming the Issue title and Pull request title).
Now when there would also be a possibility to add the 'thing' "action", I would propose to go for 'icon'.
So the Site administration setting would become '"Show view course icon" theme_boost_union | showviewcourseiconincoursemgnt'
Also in the descriptive text there could be mentioned the other two-click way to go to course view "By default, on the course management page, Moodle requires you to open the course details before you can view the course. By enabling this setting, you can add a 'View course' icon directly to the category listing on the course management page."
→
"By default, on the course management page, Moodle requires you to either open the course details or to pass through the course settings before you can click an additional UI element to view the course. By enabling this setting, you can add a 'View course' icon directly to the category listing on the course management page."

Very nit-picky but probably worth it.

lucaboesch:
Also, what about trying to use the course icon (pretend it's the link color)?

image

abias:
Hey @lucaboesch ,

thank you very much for your insights!
I fully agree that we should use a consistent terminology. I have changed the patch to use "icon" consistently.

Regarding your proposal about the icon, I think I can't change this icon and the mortarboard icon is perfectly fine.
If you look at the patch, you will see that it just uses the pix_icon i/course. And this icon is mapped to fa-graduation-cap on https://github.com/moodle/moodle/blob/master/lib/classes/output/icon_system_fontawesome.php#L217. The three-dots-circle icon is an invention of previous Moodle versions and simply does not exist in font awesome.

I have pushed an updated patch and would then merge the PR as soon as the tests have passed again.

Cheers,
Alex

Issue #428, opened 2023-10-19
Bug: Missing standard_end_of_body_html in footer


danowar2k:
_templates/theme_boost/footer.mustache_ is missing a line for the core setting **additionalhtmlfooter**:

`{{{ output.standard_end_of_body_html }}}`

danowar2k:
PR opened

danowar2k:
Nope, that won't work, going back to draft...

PR #429, opened 2023-10-19
Draft: Bugfix: Added missing standard_end_of_body_html() call in footer, solves #428

danowar2k:
Solves #428


danowar2k:
Nothing to see here...

PR #443, opened 2023-11-04
Bugfix: Show no text_to_html class element in footer fixes #442

### LINKED ISSUE #442: Bug: Footer is restricted in width because of a text_to_html div
If you want to do more in the footer, e.g. use the whole width and insert for example 
```
<div class="col-12 row">
  <div class="col-sm-3"></div>
  <div class="col-sm-3"></div>
  <div class="col-sm-3"></div>
  <div class="col-sm-3"></div>
</div>

```
because you want a four column footer spanning over the whole width, as of now that wouldn't work because the way the content is given out it is passed through a function that puts it in an additional ```<div class="html_to_text"></div>``` container which then does not span through the whole width.

lucaboesch:


CHANGES.md
```diff
@@ -6,9 +6,10 @@ Changes
 
 ### Unreleased
 
+* 2023-11-04 - Bugfix: Pass footer content without text_to_html div generation, solves #442.
```
abias, line 9: Just to be clear: This bugfix is about the _footnote_, not the _footer_. Would you mind to fix this in the changelog?

CHANGES.md
```diff
@@ -6,9 +6,10 @@ Changes
...
+* 2023-11-03 - Bugfix: Add missing cachedef strings to the language pack, solves #441.
```
abias, line 10: I think this line should not be moved within your commit. Maybe it was moved during a rebase of yours?

tests/behat/theme_boost_union_contentsettings_footer.feature
```diff
@@ -19,18 +19,27 @@ Feature: Configuring the theme_boost_union plugin for the "Footer" tab on the "C
...
     Then "#footnote" "css_element" should exist
-    And I should see "Whatever footnote" in the "#footnote" "css_element"
+    And ".text_to_html" "css_element" should not exist in the "#footnote" "css_element"
+    And I should see "Footnote" in the "#footnote" "css_element"
+    And I should not see "<span lang=\"en\" class=\"multilang\">Footnote</span>" in the "#footnote" "css_element"
+    And I should not see "FootnoteFussnote" in the "#footnote" "css_element"
     And I log out
     And I log in as "teacher1"
     And I am on "Course 1" course homepage
     Then "#footnote" "css_element" should exist
```
abias, line 36: Don't you want to add the
`And ".text_to_html" "css_element" should not exist in the "#footnote" "css_element"`
line once more here?

tests/behat/theme_boost_union_contentsettings_footer.feature
```diff
@@ -19,18 +19,27 @@ Feature: Configuring the theme_boost_union plugin for the "Footer" tab on the "C
...
-    And I should see "Whatever footnote" in the "#footnote" "css_element"
+    And I should see "Footnote" in the "#footnote" "css_element"
+    And I should not see "<span lang=\"en\" class=\"multilang\">Footnote</span>" in the "#footnote" "css_element"
+    And I should not see "FootnoteFussnote" in the "#footnote" "css_element"
     And I log out
     And I follow "Log in"
     Then "#footnote" "css_element" should exist
```
abias, line 42: Don't you want to add the
`And ".text_to_html" "css_element" should not exist in the "#footnote" "css_element"`
line once more here?

abias (changes requested):
Thank you very much, Luca, for handling this issue.
I think the change is reasonable and it's great that you check multilang tags now as well within the Behat test.

Anyway, I have added some minor comments.
As soon as you have handled them, you can directly merge the PR yourself :)

CHANGES.md
lucaboesch: That was intentional.
We do sort chronologically backwards don't we?

lucaboesch:
I've accidentally 'resolved' as closed although it is not. The meaning is to check it on the list and set the status to 'open' again.
But I haven't the proper rights therefore.

Issue #444, opened 2023-11-04
Improvement: Allow for complete removal of the "footer" (question mark in circle) button


lucaboesch:
In case you do not need it (because you have all your links elsewhere) there should be the possibility to remove the "footer"  (question mark in circle) right hand bottom button entirely.

<img width="368" alt="image" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/377279/f24faa80-ea25-4ad4-ac3e-efe0d0ec169a">

wiebkemueller-hsh:
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/6#issuecomment-1525411812

Hello Simeon, @moniNaydenov does this relate to the mentioning of Eva for PR in the issue linked above?

Regards
Wiebke

ksteitz:
We completely turned off the in our eyes disleading question mark using "theme_boost_union | scss" and support Lucas proposal for an option to turn it entirely off. 
Here is our CSS: 

/*remove question mark button for footer*/
#page-footer button {
    display: none;
    visibility: hidden;
}

Issue #6, opened 2022-03-17
Feature: A real footer


abias:
Moodle 4.x radically reduced the footer to a single floating question mark icon. After clicking on the question mark icon, users just see some links which used to be in the Moodle page footer before in Moodle 3.x

Institutions might need a real footer with
* The possibility to place reliable and directly visible links to the imprint, to data protection rules and user policies.
* The possibility to add some footer blocks to each page
* The possibility to hide / get rid of some or all standard links which Moodle places in the footer (or the question mark icon on Moodle 4.x)
* The possibility to show some images like badges in the footer of the page

Boost Campus already did this. The documentation can be found in 
https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus#4-tab-footer-layout-settings
https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus#image-area
https://github.com/moodle-an-hochschulen/moodle-theme_boost_campus#footnote.

However, there is the need to rethink footers for Boost Union.

abias:
In #20, the footnote was added to Boost Union. 
However, the other aspects remain open.

MURBASLMS:
Just being able to easily add a static footer for institution logo, links inc privacy and other things we need to display would be very good with this theme (which is asweome btw). I've always had to customise themes locally to do it properly. 

eg
<img width="1346" alt="Dashboard_and_Feature__A_real_footer_·_Issue__6_·_moodle-an-hochschulen_moodle-theme_boost_union" src="https://user-images.githubusercontent.com/83204517/232173111-752614cb-6c28-46a5-92e8-ae8359a07225.png">

wiebkemueller-hsh:
Hi @abias or @lucaboesch , 

do we get it right that this issue proposes a possiblity to get totally get rid off the of the "question-mark-footer" and not just alter its content? 

Regards&Grüße 
Wiebke und Yvonne

goggo24:
Hey all,

we've also found the question mark annoying, respectively we don't know at the moment, how to cope with contextual help. So our first solution was an adminsetting, which controls only the visibility:

![grafik](https://user-images.githubusercontent.com/16721900/234836013-2ed6081b-f6f8-4ca3-a8cd-246f4a8744d6.png)

If there is interest in this implementation, we can add an PR in the next weeks.

Regards, Eva K. (AMC)

lucaboesch:
That would be lovely, @goggo24 !
[...]

PR #464, opened 2023-11-08
Improvement: Add the option to suppress the footer button, solves #444.

lucaboesch:
I've been given the code mentioned here https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/6#issuecomment-1525411812 thanks to @goggo24 and @moniNaydenov and I have built upon that one.
Thanks heaps!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Changes
 
 ### Unreleased
 
-* 2023-12-05 - Improvement: Option to suppress footer button, solves #444.
+* 2023-12-05 - Improvement: Option to suppress footer (circle containing the question mark) button, solves #444.
 * 2023-12-01 - Bugfix: Static pages unnecessarily cleaned configured content, solves #486.
 * 2023-11-23 - Bugfix: Add background color to OAuth2 login button, solves #473.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2023-12-05 - Improvement: Option to suppress footer button, solves #444.
 * 2023-12-01 - Bugfix: Static pages unnecessarily cleaned configured content, solves #486.
 * 2023-11-23 - Bugfix: Add background color to OAuth2 login button, solves #473.
 
--- a/README.md
+++ b/README.md
@@ -379,6 +379,10 @@ In this tab there are the following settings:
 
 Whatever you add to this textarea will be displayed at the end of a page, in the footer. Refer to the setting description on the settings page for further instructions.
 
+###### Footer
+
+With these settings, you can show where to show or whether to suppress the footer (circle containing the question mark) button.
+
 #### Tab "Static pages"
 
 In this tab there are the following settings:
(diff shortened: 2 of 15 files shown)


abias (changes requested):
Hi @lucaboesch ,

many thanks for working on this contribution!
And, of course, many thanks to @goggo24 and @moniNaydenov for doing the initial implementation!

I just had a look at the patch and did some polishing and alignment in a review commit (you will see what I did if you look at the diff), but, regarding the plain functionality, I am afraid that we are not quite there yet.

The new setting differentiates between Desktop+Mobile, Mobile only, Desktop only and Completely hidden. This is great.
Hiding the button completely is done is done in the mustache template, the hiding on the particular device sizes are done with CSS. This perfectly fine as well.

But here are my questions / concerns:

* Maybe I did something wrong, but in my manual tests only the setting to completely hide the footer button (which is realized in mustache) worked. The other settings did not have any real effect for me on different screen sizes. Regardless if I used your original code or my polished code. Could you please double check these options?
* I am wondering why the CSS code to show / hide the footer button targets two different selectors `#page-footer [data-region="footer-container-popover"] .btn-footer-popover` and `#page-footer .footer-content-popover.container` differently. As I understand it, both DOM elements should be handled / shown / hidden in the same way. Could you please clarify that?
* I would like to add a `(unchanged as presented by Moodle core)` suffix to one of the setting options, similar to the "theme_boost_union | navbarcolor" setting. This way, admins should be able to identify quickly which option to choose if they do not want to change Boost core at all. The natural choice for me would be "Desktop + Mobile". But are you aware that Boost core does not show the footer button on really small devices at all? Thus, the default should be "Desktop only", shouldn't it?
* Against the same background: As Boost core does not show the question mark icon on really small devices, should we re-show it in Boost Union if the admin chooses "Desktop + Mobile" or "Mobile only"? I understand the code that this is intended (and just not working for me), but it might be worthwhile then to add this fact to the setting description.
* Against the same background: I think the default in the switch within the theme_boost_union_get_scss_to_hide_footer_button() function should return the CSS code for the default setting, not for the "None" setting as it does now.

Looking forward to your feedback!

PS: I did not touch and check the Behat tests up to now as I wanted to discuss these items first.

Thanks,
Alex

lucaboesch:
Some ``` /* stylelint-disable-line declaration-no-important */``` to add but then it is good.

abias:
After chatting to Luca 1:1 several days ago, I went ahead and tried to solve the remaining issues myself.
The cause for the issue that I did not see any effect of the settings initially was in my development environment, this is solved now.

Beyond this initial issue for my review, I have handled all other review items myself. The main challenge was to cover all possible permutations of resulting positions for the footer button, communications button and back to top button together with the sticky footer and the smart menu bottom bar. I hope that I covered everything correctly even though the CSS code has become quite lenghty now.

abias (approved):

abias:
I will merge this now, but I will not squash the commits (as I normally do) as the review changes have become quite heavy and I think it would be better to distinguish my work from @@moniNaydenov's work in the commit history.

Issue #379, opened 2023-08-27
Bug: Hide back to top button on small screens as soon as the right hand drawer is opened


abias:
On larger screens, the back to top button is move to the left as soon as the right hand drawer is opened. This is done to align it with the footer help button which is placed by Boost Core.

This is working nicely:
<img width="1190" alt="grafik" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/1092118/82388b3f-5bdc-4fe0-bff8-b3dfa66c8995">

The SCSS to realize this is here:
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/f5d593275c6ec4e8d2e446e3f82a51823e2f6e87/scss/boost_union/post.scss#L337-L340

Howerver, on smaller screens a backdrop is added as soon as the right hand drawer is opened. In this case, moving the back to top button does not make sense anymore:
<img width="873" alt="grafik" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/1092118/f523e5aa-fd38-4f8e-95ea-565a1c7b045e">

On smaller screens, the back to top button should be hidden as soon as the right hand drawer is opened

abias:
This was solved by #470

PR #470, opened 2023-11-09
Bugfix: Hide back to top button on small screens as soon as the right hand drawer is opened, solves #379.

abias:
On Moodle 4.3, the page is scrolled up automatically on smaller screens as soon as the right hand drawer is opened. Thus, this bug does not really affect Moodle 4.3.

However, I have also tested this patch on 4.2 and 4.1 (to which it will be backported, of course, during the next release) and it solves the bug there nicely.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


lucaboesch (approved):
This looks good to me.

lucaboesch:
Merging.

Issue #469, opened 2023-11-09
Bug: Bulk actions widget overlays course header image


abias:
On Moodle 4.3, the new bulk actions widget overlays the course header image (if activated and configured as stacked layout):

<img width="989" alt="grafik" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/1092118/9dc5c61c-d5ab-4364-b1a2-f01bf467daa9">

abias:
This was solved by #471

PR #471, opened 2023-11-11
Bugfix: Bulk actions widget overlaid course header image, solves #469.

abias:
Before:
grafik

After:
grafik

This was a quick fix in CSS, but it should be sufficient from my point of view.

While the "bulk actions" widget on Moodle 4.3 is the first widget where I realized this header actions area myself, the header actions area has been there before on Moodle 4 already and might be filled by course formats or plugins. Thus, it should be safe to backport this patch to pre-4.3 versions of Boost Union as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2023-11-11 - Bugfix: Bulk actions widget overlaid course header image, solves #469.
 * 2023-11-09 - Bugfix: Hide back to top button on small screens as soon as the right hand drawer is opened, solves #379.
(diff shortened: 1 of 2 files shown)


lucaboesch (approved):
I see what you are doing there, and I am willing to approve this (in fact I do) but the more proper way would be to add dynamically add a .button-secondary class to it or load button-secondary SCSS mixins, I think.

abias:
Thank you for your approval, Luca.
I understand your concerns. However, I think that we cannot rely on the fact that a button is displayed in this header actions area. I looked a little bit closer and saw that it just needs a all to $PAGE->add_header_action() to show content in this area.

That's why I just made sure that the content, whateever it is, is displayed on a white background just as Boost Core does it.

I will now rebase the branch and merge it as you have approved it.

Cheers

Issue #473, opened 2023-11-16
Bug: OAuth2 login button does not have a background image


abias:
Steps to reproduce:
* Login as admin
* On /admin/tool/oauth2/issuers.php, add an OAuth2 provider which can be used for login
* On /admin/settings.php?section=manageauths, enable the OAuth2 auth plugin
* Enable the` theme_boost_union | loginformtransparency` setting
* Upload a login background image to `theme_boost_union | loginbackgroundimage`
* Logout
* Go to the login page

Expected result:
* The OAuth2 Login button is readable

Actual result:
* The OAuth2 Login button is transparent and not readable
![grafik](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/1092118/f833a756-693b-4ea9-b28f-51aab7ecfad0)

Analysis:
* On https://github.com/moodle/moodle/blob/master/lib/templates/loginform.mustache#L168, the OAuth2 button in the login form is not given a `btn-*` class which adds a background color.
* Boost Union should override this template and add a `btn-secondary` class to these login buttons.

abias:
This was solved in #478

PR #478, opened 2023-11-23
Bugfix: Add background color to OAuth2 login button, solves #473.

wa-bea:
OAuth2 login button was transparent.
Adding btn-secondary class to this button solves the missing background color problem.

Before:

before

After:

after


abias (approved):
Thank you very much @wa-bea for providing this fix!
I have just added a review commit which adds a modification note to the overridden mustache template and will merge the PR now.

Issue #317, opened 2023-06-19
Bug: Breadcrumb navigation is partial in wrong order


stdoma:
Hello!
Breadcrumb navigation in the course is good & important! - Thanks for the option to display.

Breadcrumb navigation is not displayed correctly in all cases.
I use the MultiTopic format for courses (https://moodle.org/plugins/format_multitopic).
There, the navigation on the main page looks like this: the order of the course areas and course sections is reversed and the course itself is missing.
![Screenshot 2023-06-19 083900](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/92015860/d6c4d51e-000c-4c0a-b7f4-1bdde7371c3f)

On the individual sub-sites of the activities, the order is correct:
![Screenshot 2023-06-19 084143](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/92015860/36098d1d-7c61-48c3-9555-bfeb4dcf8900)

It's not the course format. Breadcrumb navigation is not included there.
I haven't been able to figure out the actual code yet...

stdoma:
Yesterday I took a quick look at the thing in the code.
I haven't analyzed it in detail yet.
Since the problem occurs regardless of the course format and it bothered me that the course itself also disappeared from the breabcrumbs, I modified the code a bit.
However, the solution is a bit ugly.

classes/boostnavbar.php
After the navigation is assembled I rotate the items of the object classes ("navigation_node", "breadcrumb_navigation_node").
In itself, the order of the elements is correct.
The deletion of the course can also be undone quite quickly.
After that I have the correct representation. Even if it is technically not entirely clean.
![Screenshot 2023-06-23 153243](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/92015860/a86f9ff7-b385-4399-b373-056173167a83)

If necessary, I can upload the file changes.

danowar2k:
Currently this is influenced by two settings: core/navshowcategories and boost_union/categorybreadcrumbs.

This behaviour can be experienced (pseudo code):

```
if (navshowcategories && boost_union/categorybreadcrumbs) {
   The navbar contains duplicate categories, one of which is removed because its text is the same as the others (Moodle core function)
   Left of a course's section are course categories, and right of the section are categories (basically C1 > C2 > Section > C1 > C2)
}
if (navshowcategories && !boost_union/categorybreadcrumbs) {
[...]

PR #479, opened 2023-11-23
Bugfix: Correct order for in-course breadcrumb when sections exist in…

danowar2k:
… it (First categories then sections), solves #317


danowar2k:
Huh, being in draft mode doesn't prevent the github action to run... :-/

PR #483, opened 2023-11-24
Bugfix: Correct order for in-course breadcrumb when sections exist in it (First categories then sections)

danowar2k:
Solves #317

### DIFF
--- a/tests/behat/theme_boost_union_feelsettings_navigation.feature
+++ b/tests/behat/theme_boost_union_feelsettings_navigation.feature
@@ -206,6 +206,34 @@ Feature: Configuring the theme_boost_union plugin for the "Navigation" tab on th
       | yes     | should      |
       | no      | should not  |
 
+  Scenario: Setting: Course category breadcrumbs and course sections
+    Given the following "categories" exist:
+      | name           | category | idnumber | category |
+      | Category E     | 0        | CE       | 0        |
+      | Category ED    | 1        | CED      | CE       |
+    And the following "courses" exist:
+      | fullname  | shortname | category |
+      | Course C1 | CC1       | CE       |
+      | Course C2 | CC2       | CED      |
+    And the following "course enrolments" exist:
+      | user     | course | role           |
+      | teacher1 | CC2    | editingteacher |
+    And the following config values are set as admin:
+      | config              | value     | plugin            |
+      | categorybreadcrumbs | yes       | theme_boost_union |
+    And "Course C1" has been set to one page per section
+    And "Course C2" has been set to one page per section
+    When I log in as "teacher1"
+    And I am in the second section of "Course C1" course
+    Then "Category E" "link" should exist in the ".breadcrumb" "css_element"
+    And "Topic 1" "link" should not exist in the ".breadcrumb" "css_element"
+    And I am in the second section of "Course C2" course
+    And "Category E" "link" should exist in the ".breadcrumb" "css_element"
+    And "Category ED" "link" should exist in the ".breadcrumb" "css_element"
+    And "Topic 1" "link" should exist in the ".breadcrumb" "css_element"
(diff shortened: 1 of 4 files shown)


tests/behat/theme_boost_union_feelsettings_navigation.feature
```diff
@@ -187,6 +187,34 @@ Feature: Configuring the theme_boost_union plugin for the "Navigation" tab on th
...
+      | teacher1 | CC2    | editingteacher |
+    And the following config values are set as admin:
+      | config              | value     | plugin            |
+      | categorybreadcrumbs | yes       | theme_boost_union |
+    And "Course C1" has been set to one page per section
+    And "Course C2" has been set to one page per section
+    When I log in as "teacher1"
+    And I am on section "1" page of "Course C1" course
+    Then "Category E" "link" should exist in the ".breadcrumb" "css_element"
+    And "Topic 1" "link" should not exist in the ".breadcrumb" "css_element"
```
abias, line 210: I am wondering why you are testing this case for Course C1. The teacher is not enrolled into the course, so he will end up on the enrolment page and not on the section page. On this page, the topic link is not added to the breadcrumb, that's right. But if this was really your intention, then we should add a comment about this fact to this line.

On the other hand, if this was not your intention and the teacher should have been enrolled into Course C1, then this line would fail as the topic link is displayed in the breadcrumb.

Please clarify :)

abias (changes requested):
Hi @danowar2k , many thanks for working on this bug!

The fact that the order is wrong in the breadcrumb is clearly a bug and your patch solves this bug.

I have just rebased your branch onto latest master, have generalized the Behat steps which have been introduced by you a little bit and have pushed it to your branch again.

However, I have one questions regarding your Behat scenario which I would like to ask you to answer before we merge this.

In addition to that, I have raised #593 and #594 as follow-up issues. I would be grateful if you could have a look at these as well.

abias:
@danowar2k - I do not want to bug you, but I would appreciate your answer to the question above. As soon as I have understood your intention, I will be able to finish the review and merge the PR.

danowar2k:
Whoops, completely overlooked this among the many other mails I'm getting from Boost Union on Github. I should probably change my notification settings... Sorry, will look into it

tests/behat/theme_boost_union_feelsettings_navigation.feature
danowar2k: Okay, I've looked at the code again and tried to remember...

First, I looked what a user sees when he is not enrolled (without the patch):
image

Independent of my reasoning for my PR I'd say that logically, the categories should come first, then "Einschreibeoptionen". Okay, I'd even say that before "Einschreibeoptionen" should maybe be a link to the course itself, but I'll be not picky.

But this doesn't in any way explain why I did the behat test that way. What I probably meant was to test the way the breadcrumb looks if you can view the section without being enrolled. So either as an admin or as someone with the right permission.

I'd suggest this:

* You decide if you'd like me to change the behat test to "test as admin" or to "test as user with permission to view course without being enrolled" and I'll see to it
* You decide if I should add another issue and/or behat test for the above thing with the "Einschreibeoptionen" placement (or even if you agree with me at all regarding this)

What do you think?

abias:
Thanks, @danowar2k , I think I got your point now.
I have amended the Behat scenario to test the placement of the "Enrolment options" string for non-logged-in users as well.
This is fine now from my point of view, I will merge the PR as soon as the tests are green.

Regarding your comment about the fact that "Enrolment options" string could be a link: I think this would be solved as soon as #594 would be solved. In this case, the course main page would be shown between the category name and the "Enrolment options" strings as an additional breadcrumb item. If you have some spare time, I would be grateful if you could look into that issue.

Cheers,
Alex

abias (approved):

Issue #486, opened 2023-12-01
Bug: Static pages unnecessarily clean configured content


abias:
Within Boost Union's static pages, you can add arbitrary rich text content. 
You can even add external / special content like a Google Maps map if you dare to fiddle with the HTML code within the rich text editor.

However, when the static page is output, this external / special content may not be visible as Boost Union unnecessarily cleans the configured content and drops things like iframe tags.

PR #487, opened 2023-12-01
Bugfix: Static pages unnecessarily cleaned configured content, solves #486

abias:


abias:
I will merge this myself now as the patch is a no-brainer

PR #489, opened 2023-12-04
Update emailbrandinginstructionli2 string in lang/en/theme_boost_union.php

lucaboesch:
There's no such thing as a "theme_boost_union language" setting. The ```<code>```  tags should end earlier.


abias (approved):
Well spotted! Thank you Luca!

PR #498, opened 2023-12-12
Add namespace to class smartmenu_helpers, resolves #494.

### LINKED ISSUE #494: Class smartmenu_helper should use theme_boost_union namespace
**Describe the bug**
If any other plugin or module wants to add a smartmenu_helper class that would clash

**To Reproduce**
Try to install a plugin with a smartmenu_helper class, run 
php admin/tool/behat/cli/init.php
and notice the 
Fatal error: Cannot declare class smartmenu_helper, because the name is already in use in … smartmenus/menulib.php on line 41

**Expected behavior**
If the class would have been called theme_boost_union_smartmenu_helper then it would have been all right

**Additional context**
I cloned theme_boost_union, renamed everything, e.g. to theme_luca and then tried to run Behat.

lucaboesch:


lucaboesch:
I think this is ready for peer review.

abias:
Hi @lucaboesch ,

many thanks for this PR.

I think #494 is a somehow special case as you have cloned (and modified) Boost Union and have installed this clone alongside Boost Union. This is not usual from my point of view. However, there should be no clash of both themes anyway.

Looking at your PR, I dared to make these refinements:
* I rebased it onto the latest childtheme branch (as it will be integrated to this branch as you know)
* I simply removed the leading backslash from the `smartmenu_helper` class instead of adding an explicit namespace to the class which you did. This is because we already have a `namespace theme_boost_union` statement at the very top of the relevant files.
* I reverted adding the leading backslash to the `cache` and `context_system` classes. Instead, I made sure that we have a `use` statement  at the very top of the files as well.
* I reverted your change of the `require_once` statement in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/498/files#diff-9f3bcb4b22453e357c64fd2cf3d05cc4beb7d55123c61e5a4af91a554b53ed77R29 as I really do not understand why you changed this.

As soon as the tests are green, I will merge this PR into the childtheme branch which acts as an integration buffer for the next few days as we are working on multiple child theme support improvements which should land in master all at once.

abias:
The tests are green again.
As said above, I will merge this patch now into the childtheme branch.

PR #505, opened 2023-12-18
Improvement: Shrink description_format column size in theme_boost_union_flavours table, solves #321.

abias:


lucaboesch (approved):
Looking good.
Merging.
Thanks @abias !

Issue #362, opened 2023-08-19
Release: Fix smart menu mustache linter warnings


abias:
This is a follow-up issue to #300

If you look at a recent GHA run on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/actions, you will notice that there are several mustache linter warnings which originate from the smart menus contribution.

These warnings are:

```
/home/runner/work/moodle-theme_boost_union/moodle-theme_boost_union/moodle/theme/boost_union/templates/smartmenus-cardmenu-children.mustache - WARNING: HTML Validation error, line 2: Element “li” not allowed as child of element “body” in this context. (Suppressing further errors from this subtree.) (body>    <li data-key="" class="nav-item" role="none" data-forceintomoremenu="false">     )
/home/runner/work/moodle-theme_boost_union/moodle-theme_boost_union/moodle/theme/boost_union/templates/smartmenus-cardmenu-children.mustache - WARNING: HTML Validation error, line 9: An element with “role=menuitem” must be contained in, or owned by, an element with “role=menubar” or “role=menu”. (          <a role="menuitem" class="nav-link  "                    href="https://moodle.org"                                        tabindex="-1"                                                        >     )

/home/runner/work/moodle-theme_boost_union/moodle-theme_boost_union/moodle/theme/boost_union/templates/fontawesome-iconpicker-popover.mustache - WARNING: HTML Validation error, line 4: Attribute “aria-value” not allowed on element “li” at this point. (          <li aria-value="core:book" aria-label="fa-book"><i cla)
/home/runner/work/moodle-theme_boost_union/moodle-theme_boost_union/moodle/theme/boost_union/templates/fontawesome-iconpicker-popover.mustache - INFO: HTML Validation info, line 4: Possible misuse of “aria-label”. (If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.) (          <li aria-value="core:book" aria-label="fa-book"><i cla)
/home/runner/work/moodle-theme_boost_union/moodle-theme_boost_union/moodle/theme/boost_union/templates/fontawesome-iconpicker-popover.mustache - WARNING: HTML Validation error, line 5: Attribute “aria-value” not allowed on element “li” at this point. (          <li aria-value="core:docs" aria-label="fa-info-circle"><i cla)
[...]

PR #508, opened 2023-12-21
Fix smart menu mustache linter warnings, solves #362

prasanna-lmsace:


abias (commented):
Hi @prasanna-lmsace ,

many thanks for this PR and sorry that it remained unhandled for so long!

Looking at your PR, I had to separate it into two pieces:

1. Your changes to fontawesome-popover.js and fontawesome-iconpicker-popover.mustache are fine. I just picked them and applied them to the master branch outside of this PR in d1e9c3d01ec23f4f38fc5308661ca569ff593631.

2. Your changes to templates/core/* were out of scope of the underlying issue #362. It is correct that these files contain mustache linter warnings. But they originate to the mustache files from Moodle core, they were not introduces by Boost Union. We do not fix mustache linter warnings from Moodle core in our overwritten templates as this would diverge Boost Union more than needed from Boost Core.
Looking at your changes to smartmenus-cardmenu-children.mustache (which is part of the scope in #362, I have to admit that the same argument applies here as well as this template is just a modified version of core/moremenu_children. That's why we should not fix these linter warnings there as well.

I would like to note that I cannot support your approach with your `hidemenuitemrole` thing as introducing addition logic and code just to silence Ci warnings seems a bad approach in the long run.

The only thing which remains is your change to templates/core/moremenu.mustache where you changed `{{> core/moremenu_children}}` to a local minified version of the moremenu_children template. This change would move us further away from moremenu.mustache in Boost Core. I would like to understand if there was a functional reason for this change (if yes, then please explain it in a few words and we will integrated it) or if this was just done to make the CI checker happy as well.

Thanks,
Alex

prasanna-lmsace:
Hi @abias,

Thanks for reviewing the PR.

Regarding the changes in templates/core/*, you're correct that they were made solely to address the mustache linter warnings. I appreciate your note on the approach, and I'll keep it in mind for future changes.

Best regards,
Prasanna

abias:
Hi @prasanna-lmsace ,

thanks for your swift feedback!
In this case, I will close this PR as everything which needed to be changed has already been integrated into master!

Cheers
Alex

Issue #386, opened 2023-09-13
Bug: Back to top button is missing directly after the the scroll-spy scrolled the page


abias:
Prerequisites:
* Enable Back-to-top Button in the Boost Union Settings
* Enable Scroll-Spy in the Boost Union Settings

Steps to reproduce:
* Login as teacher
* Go to one of your courses
* Scroll down
* Verify that the Back-to-top button is shown as soon as you scroll down far enough
* Enable editing mode

Expected behaviour:
* The page is scrolled down by the scroll-spy
* The Back-to-top button is directly shown again (as we are not at the top of the page)

Actual behaviour:
* The page is scrolled down by the scroll-spy
* The Back-to-top button not is directly shown again
* The Back-to-top button is directly shown again as soon as you scroll the page again manually by at least one pixel

lucaboesch:
Fixed by #514.

PR #514, opened 2023-12-24
Bugfix: Back to top button was missing directly after the scroll-spy scrolled the page, solves #386.

abias:


lucaboesch (approved):
This looks good.
Merging.

lucaboesch:
Thanks, @abias.

Issue #513, opened 2023-12-24
Tests: Use a dedicated and simple step to purge the theme cache


abias:
Throughout several Boost Union Behat scenarios, the theme cache needs to be purged after a Boost Union config has been set with `Given the following config values are set as admin`.

This is currently done this way:
```
    When I log in as "admin"
    And I navigate to "Development > Purge caches" in site administration
    And I press "Purge all caches"
```

It would be worthwhile to replace these steps in the GUI with a quicker step which just purges the theme cache with the API:

Matthias Buttgereit has already done it like this within one of his PRs, but this has not been integrated as we want to replace all locations where caches need to be purged at once:

```
    /**
     * Purges theme cache and reloads the theme
     *
     * @Given /^I purge the theme cache and reload the theme$/
     */
    public function delete_theme_cache_and_reload_theme() {
        theme_reset_all_caches();
    }
```

abias:
This was solved by #515

PR #515, opened 2023-12-24
Tests: Use a dedicated and simple step to purge the theme cache, solves #513.

abias:


lucaboesch (approved):
Good X-mas present.
I like steps improving automated testing.

Issue #516, opened 2023-12-26
Improvement: Do not add the advertisement tiles div to the frontpage if no tile is activated


abias:
During the review of #391 , it became apparent to me that up to now, this (empty) advertisement tile is added to the frontpage even if no tile is activated in Boost Union's setting:

```
<div id="themeboostunionadvtiles" class="container-fluid"><div class="row"></div></div>
```

Adding a dediated kill switch to the advertisement tiles does not make much sense as each tile can be deactivated already individually. 
It just would be good to add the `#themeboostunionadvtiles` div only if at least one tile is shown.

abias:
This was solved by #517

PR #391, opened 2023-09-19
Feature: Add frontpage slider functionality, solves #162

matthias-buttgereit:
This PR adds a slider functionality implemented with the bootstrap [carousel class](https://getbootstrap.com/docs/4.0/components/carousel/).

* Added tab "Slider" to "Content" admin settings
* Added setting to activate the slider feature
* Added setting to have up to 6 slides, each slide consisting of a ...
  * background image
  * link for clicking on that image (optional)
  * caption text being displayed on top of the image (optional)
  * content text being displayed under the caption (optional)
  * title for the image and title for the link to improve accessibility (optional)
* Added setting for where to place the slider on the frontpage
* Added settings corresponding to the bootstrap carousel [options](https://getbootstrap.com/docs/4.0/components/carousel/#options)

PR #517, opened 2023-12-26
Improvement: Do not add the advertisement tiles div to the frontpage if no tile is activated, solves #516.

abias:


abias:
I will merge this myself now without peer review as it's a no-brainer change and I want to have this change integrated as a blueprint for the review of #391.

PR #547, opened 2024-01-20
Improvement: Add a side entrance login page for local logins if the local login form is disabled on the standard login page, resolves #539.

### LINKED ISSUE #539: Improvement: Show the manual login form on request
In #490 where hiding the manual login form was implemented, @slaudel commented:

> Just one question, how does the local admin get logged in(, in case something is broken with SSO)? I would like to use this on one instance of our moodle systems, but I need to login via manual login to use my admin account.

@slaudel is right in this point. SSO interfaces can break and if this happens, the login into Moodle is broken completely. But to fix a broken SSO, the admin needs to log in. Since the Moodle login requires the login token (see https://docs.moodle.org/dev/Login_token), an approach where Moodle composes the login form itself is needed, the admin can't just create and submit an ad-hoc login form anywhere.

If we agree that this manual login is only necessary for expert users like admins in emergency cases, there are three obvious solutions:
1. Add some documentation which explains to the admin that he can simply re-enable the local login form by running `php admin/cli/cfg.php --name=loginlocalloginenable --component=theme_boost_union --set=yes`
2. Add a second login page to Boost Union, for example on /theme/boost_union/login.php. This login page would then provide the standard local login form without any bells and whistles regardless of the `loginlocalloginenable` setting.
3. Add a URL parameter like `/login/index.php?locallogin=1`to the normal login page which will show the local login form regardless of the `loginlocalloginenable` setting.

abias:


abias:
I will merge this myself now as the tests are green and the changes are quite simple.

PR #564, opened 2024-02-19
Feature: Allow administrators to change the order of login items on the login page without using CSS or touching the mustache template, resolves #504.

### LINKED ISSUE #504: Allow administrators to change the order of login items on the login page without using CSS or touching the mustache template
For our login page, we adjusted the order of (login) options/elements in the Moodle Core login mustache template.

First we want to show the login instructions and then the list of IdPs for SSO. Then the login form for local registration and finally the guest login button.
bildschirmfoto-2023-09-13-um-08-29-32

Unfortunately, the issues #473 and #490 cause our customization in core to be ignored and the default order to be used.

Hence my suggestion: It should be possible for administrators to configure the order of items in the menu without having to resort to “hacks” or CSS.

abias:


lucaboesch:
This is looking good.
I tested it by reverting the order of local login and guest login button.
Merging.

PR #603, opened 2024-03-22
Bugfix: Smart menu divider does not work for user menu submenus, resolves #537

### LINKED ISSUE #537: Bug: Adding a smart menu divider does not work for user menu submenus
Steps to reproduce:
* Create a smart menu which is set up to be displayed in the main navigation and the user menu
* Add the following items to this smart menu:
 * A static menu item with an arbitrary content
 * A heading menu item with the title `###` which will result in the menu item being displayed as a divider
 * Another static menu item with an arbitrary content
* Have a look at the resulting menus

Expected result:
* In the main navigation, there is a divider shown before the first and the second static menu item
* In the user menu, there is a divider shown before the first and the second static menu item

Actual result:
* In the main navigation, there is a divider shown before the first and the second static menu item
* In the user menu, there isn't any divider shown before the first and the second static menu item

Screenshots:
grafik
grafik

prasanna-lmsace:


abias:
Many thanks for this PR, @prasanna-lmsace !
I just reviewed your pull request and added some more changes in a "Review changes" commit.

As soon as the tests are green again, I will merge it.

abias (approved):

abias:
The tests have passed (the codechecker just fails because of https://github.com/moodlehq/moodle-cs/issues/142 which is out of our scope)

PR #609, opened 2024-03-27
Upgrade: Boost Union settings were moved to an admin settings category of its own to support the new theme chooser on Moodle 4.4, resolves #482.

### LINKED ISSUE #482: Upgrade: Boost Union unusable in the new site administration in Moodle 4.4
Dear developers,
since today there is a new structure in the site administration of Moodle 4.4.

The design settings now can be opened with a settings button in the design chooser. The settings for Boost Union open only an error page "Bereichsfehler!"

Ralf

abias:
This PR ships with a commended Behat test in tests/behat/theme_boost_union_general.feature which cannot be uncommented before Moodle 4.4

I will create a follow-up issue to uncomment it on 4.4.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ Changes
 
 ### Unreleased
 
+* 2024-03-25 - Upgrade: Boost Union settings were moved to an admin settings category of its own to support the new theme chooser on Moodle 4.4, resolves #482.
+               Please note: This change is backported to Moodle 4.3 to 4.1.
 * 2024-03-22 - Upgrade: Migrate the before_standard_html_head() function to the new hook callback on Moodle 4.4, resolves #604.
 
 ### v4.3-r10
--- a/settings.php
+++ b/settings.php
@@ -37,16 +37,18 @@ if ($hassiteconfig || has_capability('theme/boost_union:configure', context_syst
     // stupid "Too much data passed as arguments to js_call_amd..." debugging message if we would
     // pack all settings onto just one settings page.
     // To achieve this goal, we create a custom admin settings category and fill it with several settings pages.
-    // However, there is still the $settings variable which is expected by Moodle coreto be filled with the theme
-    // settings and which is automatically added to the admin settings tree in one settings page.
-    // To avoid that there appears an empty "Boost Union" settings page near our own custom settings category,
-    // we set $settings to null.
 
-    // Avoid that the theme settings page is auto-created.
-    $settings = null;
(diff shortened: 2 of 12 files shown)


abias:
I merge this myself without peer review as it's a no-brainer and tests have passed

lucaboesch:
I think I still see

image

on a Moodle main branch with Boost Union.

lucaboesch:
Please ignore.
All good.

I've not been to a fully updated master branch of Boost Union.

lucaboesch:
The only thing I see, though, is that Boost Union now has an entry both in the Theme selector page (admin/themeselector.php) and as well creates admin subcategories under "Appearence" (admin/search.php#linkappearance). 
image
I think the latter (creating admin subcategories under "Appearence") could/should be suppressed for 4.4 onwards.
That could maybe be accomplished by a version comparison switch or similar.

monnardj:
Also, if you edit boost_union theme settings through the cog icon (via "Appearance/Themes"), they all appear on a single page, without tabs, whereas Boost has two tabs (as in previous Moodle versions). Note also that both settings pages have a different title : "Boost" vs "Category:Boost Union".

boost_union

boost

abias:
Thank you both for your comments, @lucaboesch and @monnardj ,

your findings are both right.

1. On the Moodle 4.4 theme selector admin/themeselector.php, Boost Union is listed. This is necessary as it's the way how the admin sets the active theme. This is more or less the same way as it has always worked in Moodle, the selector UI has just been modified.

2. The theme selector links to each theme's settings page. This is a hardcoded behaviour in Moodle core in 4.4 and I haven't found a way to change this (except the ugly approach to hide Boost Union's cog icon on this page with CSS). However, Boost Union does not only have one settings page (like Boost Core or Classic), it has 6 settings pages and there may be more in the future. The theme selector is not able to support that. So the only solution which I could think of was to link from the theme selector page to the Boost Union settings category which shows all of Boost Unions settings. This settings category page is also a standard feature of Moodle - it is just not pretty.

3. Against this background, it is necessary from my point of view to keep the Boost Union setting admin categories on the "Appearance" page.

If you have any ideas how we can do better, don't hesitate to submit an issue or a PR.

Cheers,
Alex

abias:
Ah, I see what you mean now!
The behaviour which I described above is the behaviour on Moodle 4.4dev.
But the behaviour on 4.3 is now inferior than before. 
I will try to improve this and submit a new PR.

abias:
@lucaboesch @monnardj - please see my follow-up PR on #612

PR #612, opened 2024-03-30
Upgrade: Boost Union settings were moved to an admin settings category of its own to support the new theme chooser on Moodle 4.4, refines #482

abias:
I refined the PR #609.

There is now a Boost Union settings overview page which looks like this:
grafik

On Moodle 4.4dev, it is linked from the theme selector:
grafik

and the Appearance page only has the new Boost Union settings category:
grafik

On Moodle 4.3, it is linked from the "Boost Union" link which is still auto-generated by Moodle core:
grafik

I hope that this is a good compromise between a forward-facing fix for Moodle 4.4 and a not-so-big chance for 4.3 and older Moodle releases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,8 @@ Changes
 * 2024-04-01 - Improvement: Enhance the activitynavigation setting description to cover section navigation as well, resolves #536.
 * 2024-03-30 - Bugfix: Smart menu divider did not work for user menu submenus, resolves #537.
 * 2024-03-25 - Upgrade: Boost Union settings were moved to an admin settings category of its own to support the new theme chooser on Moodle 4.4, resolves #482.
-               Please note: This change is backported to Moodle 4.3 to 4.1.
+               Please note: This change is backported to Moodle 4.3 to 4.1 as well.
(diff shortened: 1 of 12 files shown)


abias:
The tests have passed (the codechecker just fails because of https://github.com/moodlehq/moodle-cs/issues/142 which is out of our scope).

@lucaboesch - may I ask you to have a look at this PR if you have time?

lucaboesch:
This page looks brilliant, @abias.
I understand it might make Behat testing easier if the entries still would exist in 4.4 but I think they should be omitted and the Behat tests rewritten to work with both version.
Here's how the adding to the settings could be omitted given the Theme selector page already exists: https://github.com/lucaboesch/moodle-theme_boost_union/tree/issue-482

abias:
Based on an internal discussion, I have added another panel which links to the settings category page:

grafik

Beyond that, my gut feeling is that we should integrate this PR now and postpone discussions how to further improve the settings structure and settings navigation into follow-up issues.

abias:
This PR has been discussed internally and is merged now.
A follow-up issue has been created on #621

PR #615, opened 2024-03-31
Bugfix: In smart menus, the search for cohorts in restrict visibility by cohorts didn't work for more than 25 cohorts, resolves #462.

### LINKED ISSUE #462: Smart Menus: search of cohort in restrict visibility by cohorts doesn't work properly
We have 5000+ cohorts in our Moodle. When I've tried to "Restrict visibility by cohorts" for one of the smart menu the list "By Cohort" shows only first 25 cohorts from the system cohorts and search only in this list of 25.
 It looks like it doesn't perform query to the server to fetch other cohorts according to the search query. 

Moodle 4.1.6+, Postgres 13.11.

abias:


abias:
Thank you @cjbsdigitallearning for confirming the fix!
I will merge the PR then.

PR #641, opened 2024-04-27
BugFix: Tab-Index for Smart Menus, resolves #619

### LINKED ISSUE #619: Tab-Index for Smart Menus
**Describe the bug**
So, some of our clients put a very (very) high focus on accessibilty.
In one of those projects we're using smart menus.

BUT, if I use the Tabulator to navigate the page, it moves quickly to "Dashboard" (the moodle core main navigation) but from there it DOES NOT move on the the smart menu entries (but goes to the content instead).

This is not expected behaviour, since we would like to navigate through all main navigation entries at once.

As far as we tested, smart menu entries are more or less the last thing on the page, when using Tabulator...

**This actually has a very high priority for us - and is hopefully an easy fix??**

**To Reproduce**
Just create a smart menu entry and try navigating to it using the Tabulator

**Expected behavior**
All main navigation entries (incl. those injected by smartmenus) can be navigated one after the other

prasanna-lmsace:


abias:
Hi @prasanna-lmsace ,

many thanks for working on this contribution.

Based on the discussion in the underlying issue #619, I will close this PR as "Won't do" without reviewing it further.
Thanks again for your time anyway!

Cheers,
Alex

Issue #152, opened 2022-11-29
Fix mustache shortcomings from Boost Core


abias:
There are some shortcomings in the mustache templates which have been adopted from Boost core to Boost Union.

On the one hand, we should stick as closely to Boost core as possible.
On the other hand, these shortcomings require to mute the moodle-plugin-ci mustache checker (see https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/master/.github/workflows/moodle-plugin-ci.yml#L95) to avoid that this check fails again and again.

To allow the mustache tests to run properly, it would be worthwhile to fix the mustache shortcomings from Boost Core in Boost Union finally.

abias:
The only remaining mustache shortcoming in Boost Union is this one:

```
(#1) HTML Validation info, line 2: Consider adding a “lang” attribute to the “html” start tag to declare the language of this document. (TYPE html><html ><head) 
```

This issue is highlighted for these three templates:
* theme/boost_union/templates/columns2.mustache
* theme/boost_union/templates/drawers.mustache
* theme/boost_union/templates/login.mustache

However, we won't be able to solve this easily unfortunately as we would have to diverge from the way how Moodle Core composes its HTML page from multiple mustache templates and adds the language from the output.htmlattributes template context.

wiebkemueller-hsh:
needs status on hold to be waited

PR #669, opened 2024-06-05
Enquire mustache lint errors.

lucaboesch:
Super small thing which IMHO deserves neither an issue nor a mention.

It's that one by the way
image

### DIFF
--- a/templates/block_recentlyaccesseditems/view-cards.mustache
+++ b/templates/block_recentlyaccesseditems/view-cards.mustache
@@ -27,7 +27,7 @@
                 "courseid": 2,
                 "coursename": "Course",
                 "courseviewurl": "https://moodlesite/course/view.php?id=2",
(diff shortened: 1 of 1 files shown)


abias (commented):
Hi @lucaboesch ,

many thanks for proposing a fix in this PR!

You are right, these mustache warnings should be fixed.
However, they are not introduced by Boost Union, they originate back to the original mustache templates from Boost core which have been copied to and modified in Boost Union.

If you look at https://moodle.org/plugins/theme_boost_union/v4.3-r13/32091?smurf=html#mustache, you will see several more of these mustache warnings. All of them share the fact that the warning is present in the Boost Union copy of the template and the original *.upstream version of the template.

We have issue #152 to monitor and ideally fix these upstream shortcomings sometime in the future. Until then, I hesitate to merge your PR as it would diverge the Boost Union template from the Boost Core template more than necessary.

Instead, I would like to encourage you to contribute this fix (and ideally even more fixes for the things which are highlighted in https://moodle.org/plugins/theme_boost_union/v4.3-r13/32091?smurf=html#mustache) to Moodle core if you have time.

Cheers,
Alex

PR #671, opened 2024-06-10
Cleanup: Introduce a dedicated Behat step to deactivate and activate debugging, resolves #670

### LINKED ISSUE #670: Cleanup: Introduce a dedicated Behat step to deactivate and activate debugging
Up to now, we are using these Behat steps in multiple places:

```
    # We deactivate debugging for a while as the Behat step would otherwise fail due to the
    # stupid 'Too much data passed as arguments to js_call_amd' debugging message which can't be avoided
    # on this settings page. This debugging message can't be avoided as we simple use too much hide_if() there.
    And the following config values are set as admin:
      | debug        | 0 |
      | debugdisplay | 0 |
```

```
    # We reactivate debugging again.
    And the following config values are set as admin:
      | debug        | 32767 |
      | debugdisplay | 1     |
```

These should be compacted into custom steps. In addition to that, these steps should be added to the 'Look' page as they will be needed there sooner or later based on the huge amount of settings (in fact, on 4.1, they are needed already now - see https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/actions/runs/9436868087/job/25992037521)

abias:


abias:
I will merge this myself as tests are green and the patch is a no-brainer

Issue #694, opened 2024-07-17
Accessibility issue smartmenus


create-martin:
**Describe the bug**
Using smart menus, the aria info aria-current="true" is not set correctly.
So they fail in standard accessibility tests.

Since we have a lot of clients, that need to be as accessible as possible (there's actually some laws coming into effect in 2025 in Europe) this is quite troublesome.

**To Reproduce**
Just create smartmenu --> e.g. just put a static page/dashbaord behind the link. aria-current is always on "dashbaord"

**Expected behavior**
having aria-current "moved" to correct place

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

wiebkemueller-hsh:
Hello @create-martin 

thank you for highlighting this and also the legal implications there are in EU law. We have two questions:

- Is the behaviour you are expecting only "behind the scenes" (the aria info aria-current="true") or is this also tied to a visual recognition to be made by the user? We've been talking about that in a ticket that should (if yes) then be linked to this issue.

- You seem to already have had a very good insight in the details of the problem. Are you in any way able to provide a PR that could be integrated? We're still running short on DEV resources and help would be very much appreciated!

Regards on behalf of the DEV team!
Wiebke

create-martin:
Hi Wiebke,

I was on vacation, hence the delay in answering.
As for the first question: to pass a11y tests, the "behind-the-scences" is probably enough. From a user experience point of view, I would very much like it, if the current menu option can also be visibly different from the other ones (using core, that would just be the underline hightlight; everything else can be done via theming anyway I suspect? is there a CSS class for it? or one could probably use the correctly-set-aria-label for it???). 

As for developing ressources. I will check, please give me a couple of days to find out

abias:
Hi all,

as I have just commented in #715 and #620:

The (visual) highlighting of the active smart menu node will be fixed by #648 soon.
The change will be backported to Moodle 4.4 and 4.3, but not to 4.1.
With this patch, the active node will also get the `aria-current="true"` attribute.

@create-martin , I would appreciate if you could verify (after #648 is integrated) that the behaviour matches your expectations. If yes, you can close this issue. If no, please let us know what is still missing.

Cheers,
Alex

abias:
@create-martin 
#648 has been integrated and I have verified that the active node gets the `aria-current="true"` attribute.

Against this background I close this issue for now.

Cheers,
Alex

Issue #620, opened 2024-04-04
Wrong highlighting of active menu item in the main menu


HsH-Wolf:
**Describe the bug**
If I place a static page like help.php in the main menu and then klick on this menu item, "home" is marked as active menu item instead of the actually visited page. (referres to #384 )

**To Reproduce**
- as an admin create a new menu item in the main menu (e.g. as custom menu item via theme settings or as static inline smart menu item)
- after saving these settings go to dashboard (now dashboard is marked as active menu item)
- now go to the new menu item (now home is marked as active menu item)

<img width="849" alt="bug-highlight" src="https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/assets/56588802/73202d60-1a5d-4453-b50f-db982ad957f8">

kaaermes:
is not fixed!

as you can see in this screenshot: the active page ist "smart help" but "home" ist highlighted instead

<img width="880" alt="Bildschirmfoto 2024-10-08 um 13 04 08" src="https://github.com/user-attachments/assets/38670ea9-5203-48c2-9aef-bfa4a431cb0e">

abias:
Hi all,

I had to see that this issue was solved already upstream in Moodle Core in https://tracker.moodle.org/browse/MDL-77732 for Moodle 4.3 and upwards. And this issue here would have been void if we would have merged PR #648 earlier which adopts these changes to Boost Union. 

#648 will now be integrated soon and will be backported to Moodle 4.4 and 4.3.
It will _not_ be backported to 4.1 due as it would require more work. This means that this issue will remain unfixed on 4.1.

Issue #384, opened 2023-09-07
Fixing missing active class in menu items made with Boost's custommenuitems


haw-fachadmin:
In Boost theme, there's an annoying issue: the links of menu items made with Boost's custommenuitems field in theme settings don't get an `active` class assigned to them.
_Notice: This matters if you have multiple custom horizontal menu items instead of putting them into a dropdown._

Would be great to get that fixed in Boost Union!

wiebkemueller-hsh:
Hello HAW-Fachadmin, 

thank you for posting this. I believe there is also the necessity to discuss this regarding the smart menu feature inlcuded in Boost Union. Menu items created with the smart menu also do not get an active class, but this might be for a reason. 

E.g. I can create a external link, in this case an active link is of no use. 
E.g. I can create a menu with submenus that are only displayed after clicking on the top categorie of the menu.

I marked this "discussion needed". 

Regards from H
Wiebke

wiebkemueller-hsh:
assigned to yvonne for requirements check

whuml:
Identified Issue as Boost Union Bug:
moodles custommenuitems "activ" (underline) works fine, as long, as "Boost Union Smartmenue" is not activated.
Activating or unhiding (show) "Smart Menus" causes that active custommenuitems are not underlined anymore.
Smartmenueitems are not underlined as activ as well/ in gerneral. 
Instead "home" is underlined as active :-(

wiebkemueller-hsh:
Hello @haw-fachadmin,

we'll close this ticket. The behaviour you described seems to be fixed in Boost (just verifyed on the official mount orange school testsystem). There are other issues related to the missing highlighting of the smart menu items that are being discussed in #715 
#620 .

Regards on behalf of the BU-team!
Wiebke

abias:
Hi all,

I had to see that this issue was solved already upstream in Moodle Core in https://tracker.moodle.org/browse/MDL-77732 for Moodle 4.3 and upwards. And this issue here would have been void if we would have merged PR #648 earlier which adopts these changes to Boost Union. 

#648 will now be integrated soon and will be backported to Moodle 4.4 and 4.3.
It will _not_ be backported to 4.1 due as it would require more work. This means that this issue will remain unfixed on 4.1.

PR #648, opened 2024-05-03
Adopt changes in /lib/classes/navigation/output/primary.php for Moodle 4.3, resolves #436

prasanna-lmsace:

PR #706, opened 2024-08-26
Bugfix: set active navigation node when smartmenus are enabled, resoves #620, #384 and #694.

Menrath:
Another summer time _theme_boost_union_ fix.

This targets at solving the issues #620, #384 and #694.

In the file `classes/output/navigation/primary.php` when  _smartmenus_ were enabled the `parent:export_for_template` does not get called ( parent class is `core\navigation\output\primary`).

**Cause of the bug:** The menu with the smartmenu nodes are just merged with the primary navigation and the custommenus via the PHP function `array_merge`. 

**Fix:** Now the smartmenus and the custommenus are merged via array_merge and later merged with the primary menu nodes via the function `merge_primary_and_custom`. This function checks for the active node, the same way it happens when smartmenus are disabled.

Maybe this should also be adjusted when the variable `$mobileprimarynav` gets set, too.

**To-Do:**
- [x] Write new acceptance tests that cover the flag

**At reviewers:**
There is still a bug when clicking on external static link in the smartmenu and using the browsers return function, that multiple nodes (the actual active one and the external one) are shown as active. But I don't have any idea how to fix this edge case.

### DIFF
--- a/tests/behat/theme_boost_union_smartmenusettings_menus_presentation.feature
+++ b/tests/behat/theme_boost_union_smartmenusettings_menus_presentation.feature
@@ -424,3 +424,70 @@ Feature: Configuring the theme_boost_union plugin on the "Smart menus" page, app
       | 1        | 0        | 0        | after     | before    | after     |
       | 0        | 0        | 2        | before    | after     | after     |
       | 0        | 1        | 2        | before    | after     | before    |
+
+  @javascript
+  Scenario: Smartmenu: Menus: Presentation - Verify that the correct menu item is displayed as active when viewing the main menu item's page.
+    Given I log in as "admin"
+    And I create smart menu with the following fields to these values:
+      | Title             | Quick links |
+      | Menu location(s)  | Main        |
+      | Menu mode         | Inline      |
+    And I set "Quick links" smart menu items with the following fields to these values:
+      | Title          | Test node                           |
+      | Menu item type | Static                              |
+      | URL            | /admin/tool/dataprivacy/summary.php |
+      | CSS class      | testnode01                          |
+    When I am on site homepage
+    Then the "class" attribute of ".primary-navigation [data-key='home'] a" "css_element" should contain "active"
+    And the "class" attribute of ".primary-navigation .testnode01 a" "css_element" should not contain "active"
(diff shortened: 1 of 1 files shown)


Menrath:
Something I am not completely happy with concerning the tests which is way I select the smartmenu top level menu items: `.primary-navigation [data-orgposition='5']`.

There seems to be no way to set the data-key for the smartmenus. Maybe one should be auto-generated by the title/text. Or is there a smarter more robust solution concerning the tests?

wiebkemueller-hsh:
Hello @Menrath as discussed earlier, could you have a look at this issue while at it? There might be synergies. https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/694

Menrath:
The behat tests now fail for the newest MOODLE_404_STABLE, but I think the issue is not related with this PR.

abias:
Hi @Menrath ,

many thanks for working on this PR for this set of issues.

Unfortunately, I have to say that these issues are already addressed by #648 which adopts an upstream change from Moodle core (https://tracker.moodle.org/browse/MDL-77732) which was affected by the same highlighting issues (in the Moodle core custom menu).

#648 will be merged soon and the issues are solved as a first step. This means that your PR is not needed anymore in its full extend, I am really sorry for the time which you have spent.

However you raised two additional items:

> There is still a bug when clicking on external static link in the smartmenu and using the browsers return function, that multiple nodes (the actual active one and the external one) are shown as active. But I don't have any idea how to fix this edge case.

I cannot reproduce that finding with the patch from #648.
I would appreciate if if you could re-test as soon as #648 is merged and submit a new issue if the problem persists for you.

> Write new acceptance tests that cover the flag

Your PR does not only fix the bug, but also adds Behat tests to cover the fix. I really appreciate that.
May I ask you to change this PR in a way that just the Behat test is added and works with the patch from #648? That would be awesome.
And it would be even more awesome if you could write a second Behat test which adds a custom menu (on admin/settings.php?section=themesettingsadvanced) instead of a smart menu and verifies that Boost Union does not break the highlighting of the custom menu as well.

Many thanks in advance,
Alex

abias:
Update: I went ahead and modified the Behat tests myself and will update this PR here shortly.

abias:
I rebase and force-pushed the branch and will merge it as soon as the tests are green again.

PR #710, opened 2024-09-05
Feature: Add selectable SCSS snippets

Menrath:
This is a result from a DevCamp session at the 2024 https://moodlemootdach.org

Thanks for @abias, @bartdenhoed and the others!

**Note:** All commits in this PR should definitely get squashed before merging.

**Description:**
TODO

**What is still missing:**
- [x] Unit-Tests, e.g. parsing the SCSS top level header, adding new snippets to the DB, sorting, etc., fetch enabled SCSS snippets.
- [x] Verify Copyright
- [x] Improve PHPdocs
- [x] Think whether we need to sanitze the Meta values in the SCSS top comments (e.g., via the HTMLPurifier library)
- [x] Fix all ToDos in the code
- [x] Add a modal popup for showing the details of the snippet
- [x] Extend the resolve_page_url() function for Behat which is added within #767
- [x] Add some Behat tests for the general functionality
- [x] Allow adding snippets via filepicker, including using the snippets downloaded from https://github.com/moodle-an-hochschulen/moodle-theme_boost_union_snippets 
- [x] Fix sortorder gaps (in the database) when snippets are deleted
- [x] Finalize the built-in snippets (code, description, screenshot)
- [x] Finalize the https://github.com/moodle-an-hochschulen/moodle-theme_boost_union_snippets repo (and remove the wip tag)
- [x] Add 'Tested on' metadata to snippets

**Additional features / aspects:**
- [ ] Remove the CSS header comments from the SCSS file when it is added to the SCSS stack as they would otherwise be shipped to the browser (which is an unnecessary overhead).
- [ ] Add a restriction to each snippet to only include it if a particular plugin is installed
- [ ] Add a restriction to each snippet to only include it if Moodle core version is lower / greater than a given value
- [ ] Build concept and software support to retrieve SCSS snippets from marketplace
- [ ] Add snippets from a private Github repo
- [ ] Add snippets in child theme
- [ ] Add inline search to the snippets table
- [ ] Allow the admin to automatically activate / deactivate the snippets based on time constraints
- [ ] Allow snippets which set pre SCSS variables
- [x] Try to ship the updated SCSS directly after reloading the page and not just with the subsequent page load
- [x] Fix redirect to /theme/boost_union/snippets/overview.php when the settings tab is saved
- [ ] Do not purge the cache after each change on the overview page, provide an "Apply changes" button instead
- [x] Show the CSS code in the information modal, just for information purposes

**To Discuss:**
- [x] Snippets filename: provide recommendation for either using underscores or dashes.


abias:
Hi @Menrath ,

many thanks for taking the time to submit this PR after the Devcamp!
I just went ahead an squashed the commits in your branch and amended the list of possible additional features based on the results of the Devcamp as well.

As soon as I have time, I will then review the PR.
Please stay tuned.

Cheers,
Alex

Menrath (commented):
Thanks! Those improvements all look good to me!

Menrath:
@abias I started some drafts on manually uploading snippets: Two possible solutions: just use a filepicker and add a delete action to the table, or use the filemanager and it's builtin "management" of uploaded files. I would favour the first, as the UI could be more compact. Maybe also putting the pilepicker in a modal to save UI space?

abias:
> I started some drafts on manually uploading snippets: Two possible solutions: just use a filepicker and add a delete action to the table, or use the filemanager and it's builtin "management" of uploaded files. I would favour the first, as the UI could be more compact. Maybe also putting the pilepicker in a modal to save UI space?

@Menrath - Thanks for asking.
Frankly, I would favour the second one as all the management functionality is already there in the filepicker and it matches the way how BU manages other uploaded files.

I just went ahead and pushed some additional commits which introduce a settings tab on the CSS snippets page on which you will find a (already functional) kill switch for the built-in snippets and a (currently not yet working) filepicker for uploaded snippets. I would appreciate if you could go ahead and add the necessary logic to load and use the uploaded snippets on this basis.

Menrath:
I continued working on this, but I am not quite happy with the upload mechanism of additional snippets.

I already implemented a parser that handles zip files or direct files, checks if they are valid snippets and adds them to a filearea of the plugin and the database.

Managing the "available" and "enabled" snippets on two different pages is maybe somehow more complicated than it needs to be.

@abias I know we had some arguments for this design, but I cant quite recall them :/ What do you think about having an extra filepicker to "upload" snippets and handle the enabling and deletion completely via the table?

The current admin_setting_configstoredfile for the uploaded snippets gets just too complex, if the same snippet could be uploaded in different ways at the same time. I tried to immediately extract all valid snippets in the zip and show them there, but this is also maybe somehow strange for the users, to upload a file, and see parts of it extracted.

Is it really important to keep snippets in the database that are deleted and "remembering" their settings?

Menrath:
There are still some little bugs, e.g. the sorting when only <one> uploaded snippet is available and built-in ones are disabled.

And complexity/inconsitencies: built-in snippets are always present in the database, no matter whether they are active, and uploaded ones are removed from the DB currently when they are not available.

We can merge this as is for now, but I would propose an improvement to get rid of the "Settings" tab and just add the table and always show the built-in ones. And be able to upload new ones within a Pop-Up File Picker and allow to delete those right in place.

abias:
After some more work, this PR is going to be merged quite soon.

To track the things which were not implemented up to now, I have created a follow-up issue on 
https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/989

abias (approved):

PR #720, opened 2024-10-01
Fix removed /cache/classes/loaders.php, resolves #708

### LINKED ISSUE #708: MDL-82158 removed /cache/classes/loaders.php, must be addressed otherwise.
Since MDL-82158, the file /cache/classes/loaders.php required by classes/cache/loader.php does not exist any more.
The functionalities have to be embedded otherwise.

ziegenberg:
Also, `cache_application` was renamed to `application_cache`.

### DIFF
--- a/classes/cache/loader.php
+++ b/classes/cache/loader.php
@@ -26,12 +26,12 @@ namespace theme_boost_union\cache;
 
 defined('MOODLE_INTERNAL') || die();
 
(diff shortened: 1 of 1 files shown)


abias:
Thank you @ziegenberg for this PR!
Your solution is spot-on. I have cherry-picked it into the upgrade-405 branch which is the basis for our Moodle 4.5 development.

PR #763, opened 2024-11-19
Fix thirdparty icon size in 4.4 courseindex drawer, resolves #755

### LINKED ISSUE #755: Backport: Port the activity icon sizes in the nav drawer back to stable versions
We have noticed recently that the activity icon sizes in the nav drawer can appear incorrectly with some third-party activity icons.

In 4.5, this glitch is already solved by https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/commit/d0822df029fbf89402ac4e6f816073902c07c3b8. The activity icon sizes from this patch should be backported to stable versions.

christianwolters:
This PR provides explicit width and height for icons in the courseindex drawer for Moodle 4.4.

As mentioned in the issue, this is needed if icons exceed a specific size as seen with some thridparty plugins.

To provide consitency with 4.4. styling, 24px were used as values.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### v4.4-r6
 
+* 2024-11-19 - Bugfix: Port the activity icon sizes in the nav drawer back to stable versions, resolves #755.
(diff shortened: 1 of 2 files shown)


abias:
Thank you very much for this fix, @christianwolters .
The change locks fine, I have just added a "resolved" suffix to the CHANGES.md entry, and will merge the PR now.

PR #764, opened 2024-11-19
Upgrade: Support more seamless upgrade to 4.5 by only requiring /cache/classes/loaders.php if it is present, resolves #708.

### LINKED ISSUE #708: MDL-82158 removed /cache/classes/loaders.php, must be addressed otherwise.

abias:


abias:
Note: This PR must not be merged into main and 4.5. It will be merged into 4.4 and backported to 4.3 and 4.1

ralf-krause:
Hi Alex,
you add some code into the next Boost Union releases v4.1-r33, 4.3-r19 and 4.4-r6. v4.2-r21 is missing because an upgrade also can be done from Moodle 4.2 to Moodle 4.5. The minimal is Moodle 4.1.2 to update to Moodle 4.5 … and this includes 4.1.2 and better, and all versions 4.2.x, 4.3.x, and 4.4.x.

I fear that this will not fix the problem completely. An admin must know that the Boost Union update to the next version is required before he starts the upgrade to Moodle 4.5.

Ralf

abias:
Hi @ralf-krause ,

I see your point.
But this upgrading issue is a problem of _already installed_ pre-4.5 plugin versions.
We made a mistake in the past and we cannot turn back time.

I can only publish new plugin releases for older Moodle versions and hope that the admins update to these latest 4.1 / 4.3 / 4.4 plugin versions _before_ they move on to Moodle 4.5. If the admins do this, they won't be affected by the upgrading issue.

Regarding 4.2, this Moodle version is out of support for more than a month now and we have stopped publishing new BU versions for 4.2 since then. Even if I would publish a new BU release for 4.2 with the fix, I would not assume that admins will install it if they do not even manage to keep their Moodle core up to date and safe.

And anyway, this issue is indeed a blocker for the 4.5 upgrade, that's right. But it does not destroy any data. As soon as the BU plugin is upgraded to 4.5, Moodle core can be upgraded without any issue.

Cheers,
Alex

PR #767, opened 2024-11-22
Behat optimisations, addresses #765.

marxjohnson:
This does a bunch of optimisations for the longest-running behat features. See the commit messages for details of what I've done. Further optimisations are possible, but I don't have time to work on this any further at the moment. Hopefully the groundwork of adding the `behat_theme_boost_union` and `behat_theme_boost_union_generator` classes will make it easy for others to make further improvements.

In [my first run of these changes](https://github.com/marxjohnson/moodle-theme_boost_union/actions/runs/11957413640) the tests completed in 2.5 hours, vs the 3.5-4 hours I was getting previously., 

[x] link your issue in the PR title, using the keyword 'resolves #ISSUE-NUMBER', e.g. 'Feature: Provide the ultimate user experience, resolves #42'
[x provide any further information that is relevant for peer review and not yet mentioned in the linked issue as a comment in the PR
[x] make sure that the 'Allow edits by maintainers' checkbox is checked when creating the PR. Otherwise, the peer reviewer would not be able to push any review changes to the PR and the communication overhead increases
[x] submit your PR in draft status to run the automated checks and review the results
[x] in case any checks fail solve the mentioned errors by pushing the corrected code to your PR-branch
[x] if all checks pass (or if you are unable to resolve the failing steps without any help of the review team), mark the PR as 'ready for review'

Issue #765, opened 2024-11-22
Behat test optimisations


marxjohnson:
This plugin has a very comprehensive suite of behat tests, which is good, especially considering the range of features it implements. 

However, the current suite takes about 4 hours to run, which presents a bit of a barrier when contributing fixes.

I captured the timings (in seconds) for each feature. Looking at the smart menu tests in particular, there is definitely scope for optimising these with generators, navigation URLs and named selectors. I will submit a branch with some improvements.

```json
{
    "theme\/boost_union\/tests\/behat\/theme_boost_union_smartmenusettings_menuitems_dynamiccourses.feature": 3099.5,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_feelsettings_blocks.feature": 3016.7,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_smartmenusettings_menuitems_rules.feature": 2833.1,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_smartmenusettings_menus_rules.feature": 2303.5,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_contentsettings_staticpages.feature": 1690.8,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_smartmenusettings_menus_presentation.feature": 1258.6,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_smartmenusettings_menuitems_presentation.feature": 1194.3,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_contentsettings_slider.feature": 1085.2,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_contentsettings_footer.feature": 784.8,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_feelsettings_navigation.feature": 486.3,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_looksettings_course.feature": 468.7,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_looksettings_sitebranding.feature": 424.3,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_looksettings_scss.feature": 408.8,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_contentsettings_advertisementtiles.feature": 367.6,
    "theme\/boost_union\/tests\/behat\/theme_boost_union_flavourssettings_caching.feature": 359.9,
[...]

PR #769, opened 2024-11-22
 Behat optimisations, addresses #765 (404 backport)

marxjohnson:
Backport of #767 to the MOODLE_404_STABLE branch.


marxjohnson:
The failing runs were due to a selenium crash, We've got a pass on each PHP version and each database so I think we're safe to say it's OK.

PR #779, opened 2024-12-06
Replace lower case Theme boost union by upper case.

lucaboesch:
Theme boost union

should be

Theme Boost Union


lucaboesch:
Trivial change for beauty

PR #799, opened 2025-01-25
Bugfix: Fix dynamic course items not updating properly based on course role assignments. resolves #749.

### LINKED ISSUE #749: Smart Menu Item Dynamic Courses
**Describe the bug**
Unenrolling from course does not remove course from smart menu

**To Reproduce**
1. Create new smart menu (I used submenu/list)
2. Add an item - I selected Dynamic, Courses, Time in Future, Role as Student
3. Enroll in one or more future courses
4. Check menu behavior
5. Unenroll from future classes.
6. Class link still listed.

**What happened**
1. I expected at least the next upcoming course to be listed.  I was not sure what would happen if there was more than one.  I think it would be helpful to be able to limit it to one course - basically I would like to be able to include the student's next upcoming training in the menu.
2. When I was enrolled in only one future course, it was listed as expected.  I then enrolled in a second course but nothing changed.  Wasn't sure how that would work anyway as it is an item, not a whole submenu.
3. I then unenrolled from the one class that was showing in the menu.  I expected the menu to then change to the next course that I was enrolled it.  But it didn't change at all.  It still showed the course that I was no longer enrolled in.

**Desktop (please complete the following information):**
 - OS: [Mac]
 - Browser [tested in chrome and brave

**Additional context**
Testing on Moodle 4.5 site.

prasanna-lmsace:
Fixed the issue where course enrollment and unenrollment now properly update the dynamic course items for users.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,9 +3,9 @@ moodle-theme_boost_union
 
 Changes
 -------
-
 ### Unreleased
 
+* 2025-02-04 - Bugfix: Fix smart menu dynamic course items not updating properly based on course role assignments, resolves #749.
 * 2025-02-04 - Bugfix: Smart menu item pointing to external site gets highlighted as active by mistake, resolves #758.
 * 2025-02-04 - Improvement: Allow changing of home URL on small devices as well, resolves #802.
 * 2025-01-31 - Improvement: Add option to include alt text for item image in smart menu cards, resolves #752.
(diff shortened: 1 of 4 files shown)


abias:
Thank you @prasanna-lmsace for fixing this reported bug so quickly!

Your fix seems to be spot-on, especially as you covered it with a Behat scenario.
I just rebased your bugfix branch onto latest main and added some homekeeping review changes. 

I will merge the PR as soon as the tests are green again.

Cheers,
Alex

PR #832, opened 2025-02-18
Correct activity icon colors, resolves #831

### LINKED ISSUE #831: Activity icon colors are not accurate
**Describe the bug**
The same problem currently exists in the core Boost theme in Moodle versions 4.4 and 4.5 and is described in https://tracker.moodle.org/browse/MDL-84306.

**To Reproduce**
1. Change theme setting `activityiconcolorassessment` to _#5B40FF_
2. Go to a course
3. Create an assignment
4. Open the assignment's page
5. Zoom the page to make the icon as bis as possible
6. Take a screenshot of the icon
7. Insert it in an image editor (e.g. GIMP)
8. Use the color picker to check which color the icon has
9. Notice the color differs significantly from the intended color.
   * Example: #7d4bff

**Expected behavior**
The color should be very near the intended color.

**Screenshots**

Image

 - Moodle Version: 4.4/4.5
 - Boost Union Version: latest 4.4/4.5
 - OS: Windows
 - Browser: Chrome

t-schroeder:
Make activity icons black before applying color filters for final color.

Looking at https://codepen.io/sosuke/pen/Pjoqqp (which I assume the CSS filter calculation is based on, same as in Moodle core up until 5.0 (see [MDL-83725](https://tracker.moodle.org/browse/MDL-83725))) I saw this paragraph:

> For this code to work well the starting color needs to be black. If your icon set isn't black you can prepend "brightness(0) saturate(100%)" to your filter property which will first turn the icon set to black.

In Moodle 4.4 (see [MDL-78284](https://tracker.moodle.org/browse/MDL-78284)) the core activity icons got replaced by new non-black ones. Therefore the CSS filters result in incorrect colors. For some colors this difference is actually noticeable, especially if you have third-party activities with icons with base color black next to core activities with icons with this non-black base color. I'm fixing this in Moodle core through [MDL-84306](https://tracker.moodle.org/browse/MDL-84306) and it is necessary to apply that same fix here.


t-schroeder:
This is how the color ended up for me when setting `activityiconcolorassessment` to _#5B40FF_ with `activityiconcolorfidelity` as _One shot (1 iteration)_:
image

abias:
Thank you, @t-schroeder , for this PR.
As I have just commented in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/831#issuecomment-2838117972, I would like to wait for final feedback by Moodle HQ if this issue will be solved upstream in Boost Core or not. We will keep this PR on hold until then.

PR #867, opened 2025-03-17
Remove unneeded sticky-top class from stickycategory, resolves #866.

### LINKED ISSUE #866: On course home, the stickycategory div does not allow button pulldown to hover over it
Note here

Image

the z-index of the category element is of such level that the menu is covered instead of overlapping.

Also the 'More ▾' menu would be covered if hovered.

lucaboesch:
I believe just removing that sticky-top class will do.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2025-03-17 - Bugfix: Remove unneeded class that would make menus appear under the category header, resolves #866.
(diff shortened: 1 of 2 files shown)


abias:
I will comment in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/866 as I do not fully agree with the proposed solution approach.

abias:
Closing after the insights in https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/866#issuecomment-2750440340

PR #871, opened 2025-03-20
Adjust position of drawers and stickcategoryheader with active smartmenu top bar, resolves #870 #879 #899

### LINKED ISSUE #870: Categories as list of boxes: category title is truncated with active smartmenu menubar
If you have a smartmenu active with the menu bar

Image

And configure categorylistingpresentation on /course/index.php to a "list of boxes",
the category will be truncated by the upper navbar

Image

### LINKED ISSUE #879: Smart Menu location 'Menu bar' prevents bottom item in course index from displaying
**Describe the bug**
If you add a SmartMenu item in the 'Menu bar' location, a user may not be able to see the last item (activity/resource/section/subsection) in the course index drawer. 

**To Reproduce**
1. Go to Boost Union settings > SmartMenu
2. Add a SmartMenu item for location 'Menu bar' and configure as desired
3. Open any course with enough content to require a scroll bar to appear in the course index
4. Check to see if the last content object (activity/resource/section/subsection) remains below the bottom of the drawer when all sections/subsections have been expanded.

**Expected behavior**
One would expect the last item in the course index to be visible regardless of the SmartMenu location 'Menu bar' being used or not

**Screenshots**
N/A

 - Moodle Version:4.5.2
 - Boost Union Version unknown

christianwolters:
see #870 , #879 and #899.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2025-04-02 - Bugfix: Adjust position of drawers and stickcategoryheader with active smartmenu top bar, resolves #870 #879.
+
 ### v4.5-r13
 
 * 2025-04-06 - Bugfix: Enrol page used modified course listing renderer from category index / site home as well, resolves #895.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Changes
 
 ### Unreleased
 
-* 2025-04-02 - Bugfix: Adjust position of drawers and stickcategoryheader with active smartmenu top bar, resolves #870 #879.
+* 2025-04-02 - Bugfix: Adjust position of drawers and stickcategoryheader with active smartmenus bars, resolves #870 #879 #899.
 
 ### v4.5-r13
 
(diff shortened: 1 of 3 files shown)


abias:
Hi @christianwolters ,

many thanks for this contribution.

I just rebased the PR onto latest master and made some minor review changes in a separate commit.

Regarding #870, I can confirm that the bug is solved by this patch.

Regarding #879, I think there is some room for improvement.
On the one hand, you have to note that on small screens, the menu bar is placed _behind_ the course index drawer and not above it anymore. But this means that the calculated height of the drawer does not fit anymore on small screens:

grafik

Additionally, I would like to point out that we have already some places where the `.boost-union-menubar` class is used as a selector, especially for handling the height of the message drawer together with the menu bar.

Would you mind to interweave your new handling for the course index with the existing handling for the message drawer and use the same breakpoints? And it would be a great benefit if you could also modify the handling of the message drawer to use the new `$boost-union-smartmenu-menubar-height` variable as well.

Many thanks,
Alex

abias (changes requested):
.

christianwolters:
Hey @abias,

thanks for pointing this out.

I''ve fixed this for small and middle-size screens (not pushed yet).

During testing, I encountered additonal bugs, when top menubar and footerbar are both enabled:
* I'll create a seperate issue for this
* but will adress this in this PR as well

PR #643, opened 2024-04-27
BugFix: Make outside-left and outside-right block regions really responsive, resolves #266

prasanna-lmsace:

PR #880, opened 2025-03-27
Improvement: Add vertical offset to the outside-left and outside-right regions, resolves #254.

### LINKED ISSUE #254: Improvement: Add vertical offset to the outside-left and outside-right regions
This is a follow-up issue of #30 / #118 

Currently, the outside-left and outside-right block regions are vertically aligned with the top of the #topofscroll element.

It would be worthwhile to have an additional admin setting which allows the admin to puth these regions further down to that they are aligned with the start of the content area.

grafik

prasanna-lmsace:


abias:
Hi @prasanna-lmsace ,

many thanks for working on this contribution and sorry that it took us so long to review this PR.

I just rebased it onto latest main and had a look at the code.
I think your solution approach is perfectly fine. I just added a "Review changes" commit with language and comment and settings refinements.

Then, when testing manually, I encountered a glitch:
As #643 has been merged in the meantime, I saw that the code adds too much space above the outside-left and outside-right regions on smaller viewports.

On larger viewports, it looked like this (which is perfect):
grafik

And on smaller viewports it looked like this:
grafik

My gut feeling was that the JS code should be changed in a way that the vertical alignment is only processed on larger viewports. On smaller viewports, it should not do anything. And when resizing from a larger to a smaller viewport, it should revert the vertical alignment and let Bootstrap do its job.

I just went ahead and added this detail to make the PR mergeable.
I will merge it as soon as the tests are green.

Cheers,
Alex

Issue #630, opened 2024-04-16
Internal note: add min width to smart menus


stefanscholz:
This is a follow up issue for smart menus and will be considered and spec'd out after the current smart menu enhancements are completed.

Idea: Introduce a min-width setting for each smart menu with options like:
- narrow (200px)
- wide (360px)

PR #888, opened 2025-03-30
 Bugfix: Unset width restrictions for dropdowns in custommenu and smartmenu, resolves #883

### LINKED ISSUE #883: Smartmenu influences styling of core custommenues
Custommenus in Moodle core have no restrictions on width:

Image

When activating Boost Union smartmenu function, this changes unexpectedly,
to a small width causing the text to break into the next line, like in this example:

Image

Expected behavior:

* Boost Union settings should not influence core behavior

Also: even for custommenu itemds that witdh is probably to small.

christianwolters:
See #883

This unsets the width restrictions from earlier smartmenu works.

With the new css core costummenus, but also dropdowns from smartmenus are rendered without any wrapping.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ moodle-theme_boost_union
 Changes
 -------
 
+### Unreleased
+
+* 2025-04-02 - Bugfix: Unset width restrictions for dropdowns in custommenu and smartmenu, resolves #883.
+
 ### v4.5-r12
 
 * 2025-03-29 - Bugfix: Remove assumption that syscontext->id = 1, resolves #627
--- a/scss/boost_union/post.scss
+++ b/scss/boost_union/post.scss
@@ -2471,6 +2471,12 @@ body.theme_boost-union-footerbuttonnone.jsenabled {
                 text-decoration: none;
             }
         }
+        &.dropdown {
+            .boost-union-moremenu.dropdown-menu {
+                max-width: unset;
+                width: max-content;
+            }
+        }
     }
     /* Dropdown carousel more menu*/
     .dropdownmoremenu {


abias (changes requested):
Hi @christianwolters ,

many thanks for this contribution.

I just rebased the PR onto latest master and decoupled it from #871 by removing the commits which you have submitted in #871 to allow us to concentrate on this single particular bugfix here.

Now, the bug report in #883 touches multiple things which we have to keep in mind when assessing a possible solution:

* Boost Union must not change Boost core behaviour unexpectedly, that's a consensus. However, Boost Union by design renders the Moodle core custom menu as smart menu (with one navigation level) if a custom menu _and_ a smart menu exists. Boost Union just determines the Moodle core custom menu nodes and the smart menu nodes which should be presented and shows them all together with the smart menu renderers. This was designed intentionally to ease the implementaiton of the smart menus. But against this background, we have to make sure that the custom menu items which are "shipped" as simple smart menu look as if they were custom menus or at least not substantially different.
* The custom menu in Moodle core does not have a maximum width at all:
grafik
From my point of view, this is a downside of the Moodle core custom menu and we do not need to keep this in Boost Union and should however surely have a max width for the custom menu _and_ the smart menu. We should just mention this on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/main/README.md#exceptions-to-our-main-design-principle
* In contrast to the pure Moodle core custom menu, the smart menu has a max width. Its value was set in a way that we avoid that the smart menu (which is left-aligned to its main menu item) hangs out of the right-hand edge of the viewport. This basically works well. However, there is https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/630 as a follow-up issue to give the admin some more control.

Looking at your solution now, I think that neither unsetting the max-width (which would then trigger the problem of letting the menu hang out of the screen) nor setting width to max-content (which would allow the content to explicitely exceed its container width if I am not wrong) is really the way to go.
My gut feeling is that we should rather stick to width: fit-content and raise the min-width of the smart menu for each breakpoint in a way that the menu has more space and looks good even with longer navigation item titles but still does not exceed the available viewport space.

If this is done right, https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/issues/630 might be obsolete in the end.

PR #896, opened 2025-04-05
Bugfix: control use of boost union coursebox, resolves #895

### LINKED ISSUE #895: Bug: Enrol page uses course listing renderer from category index / site home
With the new Boost Union course listing boxes for category index / site home pages enabled,
enrol pages are also changed.

The result is an incomplete enrol page, and also incomplete BU course cards as,
e.g. no course image, wrong formatting, reundant buttons and hidden information.

christianwolters:
See #895

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2025-04-06 - Bugfix: Enrol page used modified course listing renderer from category index / site home as well, resolves #895.
 * 2025-04-01 - Feature: Add moodle documentation smartmenu item type, resolves #657.
 
 ### v4.5-r12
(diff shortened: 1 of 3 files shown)


abias:
Hi @christianwolters ,

thank you for spotting and reporting this glitch and for proving a fix.

I agree with the general solution approach, but I made these adjustments to the PR:

* I moved the helper function from the course class to the course_renderer as it did not fit 100% into the course helper class in my point of view.
* I generalized the comments (from: "use coursebox" to "allow modification")
* I changed the URL comparision to use the $this->page->url->compare() function instead of a simple string comparison
* Most importantly, I fixed the Behat failures. These resulted from the fact that the modified renderer not only gets called by the renderered page itself but also by /course/category.ajax.php. So this must be respected in the code change as well

I think this can then be mrged as soon as the tests are green.

Cheers,
Alex

PR #945, opened 2025-05-13
Feature: Change AI course assistance placement button location, resolves #894

### LINKED ISSUE #894: Change AI course summarise button placement
**Is your feature request / improvement related to a problem? Please describe.**
Moodle 4.5 introduced a new [AI Subsystem](https://moodledev.io/docs/4.5/apis/subsystems/ai) which (among other features) can be set up to allow students to generate a summary of the text content of any activity page (see [here](https://docs.moodle.org/405/en/AI_placements#Course_Assistance_placement)). When this feature is enabled, a "Summarise" button is added above the content of each activity page:

Image

When clicked, this button opens a new drawer in the same position as the block drawer:

Image

We find this button to be quite intrusive as it takes up valuable vertical space even though it is not relevant for many users. We have already received complaints about this from some of our users.

**Describe the solution you'd like**
We would like to add an option to change the placement and look of this button using Boost Union.

I have come up with four suggestions for the new placement, which I will present here.

Suggestion 1: Make the button a header action (like the "Manage courses" and the "Create course" buttons on the "My courses" page). This moves the button to the right of the activity title, but leaves its design unchanged. This is easy to implement.

Image

Suggestion 2: Add the button to the list of additional buttons in the context header (like the "Message" button on the user profile page). This places the button directly after the activity title and changes its design to be less intrusive. This is a bit more difficult to implement.

Image

Suggestion 3: Add a new drawer toggle below the toggle for the block drawer. This makes sense semantically because the summarise button opens a drawer when it is clicked. It prevents the AI drawer from being opened while the block drawer is open, though. This could get tricky when it comes to styling.

Image

Suggestion 4: Add a new footer button above the default one (just like the accessibility support link which is already offered by Boost Union). This is easy because it reuses the functionality already implemented in Boost Union, but we might need to review the styling on small devices (it seems to interfere with the block drawer toggle).

Image

My personal favourites are 1 and 4, but I'm interested in hearing different opinions. If multiple versions are desired, we could allow admins to select one of multiple options.

**Describe alternatives you've considered**
An alternative would be to tell users who do not want to see the button to remove it using a client-side browser add-on (e.g., an ad blocker). But there might be users who are just annoyed by the buttons placement and not the button itself.

larsbonczek:
- Added a new "AI" tab on the "Feel" settings page
- Added a setting theme_boost_union/aiplacementcourseassistlocation to control the location of the AI button
- Added styles to remove padding and disable text wrapping on the AI button when it's a header action
- <del>Code is designed to be future-proof (works with upcoming changes to aiplacement_courseassist in Moodle 5.0)</del>
<del>The styles mentioned above will break in Moodle 5.0 :(</del>
- Added Behat tests


larsbonczek:
Note that this uses the function [`theme_boost_union_manipulate_hooks()`](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/832223669a3d30c1a71b801fa7bce4a3567d7d43/locallib.php#L2396) to prevent the AI button from being added in the default location when its location has been changed in the Boost Union settings. However, as is mentioned [here](https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/832223669a3d30c1a71b801fa7bce4a3567d7d43/lib.php#L853-L855), this function is executed regardless of whether Boost Union is selected as the current theme or not. This means that if Boost Union is NOT selected as the current theme and the new setting to change the AI button location is set to "headeraction", then the AI button will NOT be added in the default location (because the aiplacement_courseassist hook is overriden in `theme_boost_union_manipulate_hooks()`) and it will also NOT be added is the new location (because Boost Union is not selected as the current theme). This means that the button disappears completely.

To prevent this, I suggest overriding the aiplacement_courseassist hook with a custom callback (instead of disabling it) which checks whether Boost Union is the current theme and calls the original aiplacement_courseassist callback if it is not. Do you agree with this solution @abias? I'm not sure where to place this new callback. It doesn't make sense to me to put it in the logical place for hook callbacks (i.e., `classes/local/hook/output/after_http_headers.php`) because it's not a real hook callback registered in the `db/hooks.php` file. It's only called if the Boost Union setting is set to "headeraction", after all. What about `classes/local/hook_override/aiplacement_courseassist/hook_callbacks.php`?

christianwolters (dismissed):
## Review

### Documentation

- [x] **Commit Message** understandable and issue referenced (via resolves keyword)
- [x] **Patch author** correct
- [x] **CHANGES.md** --> **appended in review**
- [x] **README.md** appended
- [x] **Credits** appended
- [x] Appropriate **Comment quantity**
- [x] Successful **Moodle PHPDoc check**

### version.php

- [x] Checked if **$plugin->version** increment necessary (and increment done if necessary) -> **incremented in review**
- [x] **$plugin->release** untouched

### lib.php

- [x] Only **necessary functions**

### Languages

- [x] Only **english strings**
- [x] Necessary **magic strings** added (e.g. capabilities)

### Automated tests

- [x] **New functionality covered**
- [x] **No failing steps or scenarios**

### Mustache templates

- Not applicable **Example context exists**

### CSS and styles.css

- [x] **Bootstrap styles** used if possible

### Duplicated code

- [x] **Duplicated code is marked properly**

### Github action

- [x] **All green**

### Commit history and scope

- [x] Already **single commit**
- [x] Focus on **single purpose**
- [x] No **surplus files**

### Additional aspects for Boost Union

- [x] No usage of $theme->settings
- [x] Usage of isset() checks when processing plugin settings in renderer / output code --> **added in review**
- [x] Usage of admin_setting_configselect instead of admin_setting_configcheckbox
- [x] Modified mustache templates properly marked and .upstream template in place

## Review result

Hey @larsbonczek,

thank you for your contribution to Boost Union.

I made some minor changes, but the feature worked as advertised in my tests and review.

Cheers
Christian

Issue #1025, opened 2025-07-25
Support for Filters in Custom Menu


michael-milette:
Thank you for making Boost Union available to the Moodle worldwide community. The time you and your team have invested in this theme is much appreciated.

The issue I am having is that custom menu items are currently only filtered if the complete tag is on a single line. It fails if the complete tag spans multiple lines.

I would like to use filter plugins, like FilterCodes and others, in the Primary Navigation (a.k.a. Custom Menu).

**Example that works:** In this example, FilterCodes is used to create a Logout link that causes Moodle to prompt the user to confirm that they really do want to log out - so that they don't log out by mistake:

{ifloggedin}Logout|/login/logout.php{/ifloggedin}

**Example that does not work:** In this example, a "Themes" menu is added to the custom menu, but only if debugging is enabled:

{ifdev}
{menuthemes}
{/ifdev}

The FilterCodes {menuthemes} tag is replaced by a dropdown list of themes installed on the system for easy switching by developers. There are dozens of other conditional tags included with FilterCodes that can be used in hundreds of different ways.

**Describe the solution you'd like**

The problem has already been solved in Moodle 5.0+ core. For other currently supported versions of Moodle, a dev either needs to patch Moodle core or patch Boost Union manually. Given the popularity of Boost Union and filter plugins like FilterCodes, I think it is a great idea to add such support to supplement Smart Menus

**Describe alternatives you've considered**

I checked the options in Smart Menu but could not find a combination of options that would enable me to achieve the desired outcome.

**Additional Information**

I will be submitting a pull request shortly to address this issue.

Best regards,

Michael Milete

PR #1026, opened 2025-07-25
Resolves Issue #1025: Feature: Add support for filters in custom menu.

michael-milette:
Note: This patch is not required for Moodle 5.0+ as the functionality is now integrated. That said, if you prefer to keep it in for Moodle 5.0+, it won't hurt as the code includes a check for the Moodle version.

The following steps have been completed to ease the review process for the peer review team:

[X] link your issue in the PR title, using the keyword 'resolves #ISSUE-NUMBER', e.g. 'Feature: Provide the ultimate user experience, resolves #42'
[X] provide any further information that is relevant for peer review and not yet mentioned in the linked issue as a comment in the PR
[X] make sure that the 'Allow edits by maintainers' checkbox is checked when creating the PR. Otherwise, the peer reviewer would not be able to push any review changes to the PR and the communication overhead increases
[X] submit your PR in draft status to run the automated checks and review the results
[ ] in case any checks fail solve the mentioned errors by pushing the corrected code to your PR-branch
[X] if all checks pass (or if you are unable to resolve the failing steps without any help of the review team), mark the PR as 'ready for review'

Please let me know if you have any questions or concerns.

Best regards,

Michael

### DIFF
--- a/layout/drawers.php
+++ b/layout/drawers.php
@@ -128,7 +128,11 @@ if ($PAGE->has_secondary_navigation()) {
 // It includes the smart menus and menu items, for multiple locations.
 $primary = new theme_boost_union\output\navigation\primary($PAGE);
 $renderer = $PAGE->get_renderer('core');
+if (!empty($oldmenu = $CFG->custommenuitems) && $CFG->branch < 500) {
+    $CFG->custommenuitems = format_string($CFG->custommenuitems);
+}
 $primarymenu = $primary->export_for_template($renderer);
+$CFG->custommenuitems = $oldmenu;
 
 // Add special class selectors to improve the Smart menus SCSS selectors.
 if (isset($primarymenu['includesmartmenu']) && $primarymenu['includesmartmenu'] == true) {


abias:
Hi @michael-milette ,

many thanks for submitting this PR!

So this is a cross-port of MDL-63219 which has been integrated into Moodle 5.0.
I see that there is a backport request on MDL-84559 as well, but this is still pending. Thus, it indeed would be interesting to backport this into Boost Union 4.5 if it would help the community.

Looking at your patch, I think it is quite straightforward. However, I would like to raise three change requests:

1. As you surely now, the main design principle of Boost Union it that it does not change anything to Moodle at first and simply behaves as Boost from Moodle core does. Every change has to come with an admin setting (or it has to be added to https://github.com/moodle-an-hochschulen/moodle-theme_boost_union?tab=readme-ov-file#exceptions-to-our-main-design-principle which should only be rare edge cases).
Against this background, may I ask you to implement an admin setting which controls the added behaviour? This setting would be added to the "Feel -> Navigation -> Primary navigation" settings section and would basically be the same kill switch as $CFG->navfilter which was added to Moodle core.

2. Additionally, I would appreciate if you could add a Behat test which covers this setting and its results.
Ideally, the test would work with a filter which is contained in Moodle core. If this is not possible at all, you can add the installation of filter_filtercodes into .github/workflows/moodle-plugin-ci.yml

3. Looking at the changes of MDL-63219 in Moodle core, I am wondering why you process explicitely only string filters there, but in this patch here you just use format_string(). You are much more experienced with filters than I am, so I am quite sure that this patch here should be fine as well. But I would appreciate if you could explain the difference to us.

Many thanks in advance,
Alex

------

Note to myself:
When this PR is integrated, a follow-up issue should be created to evaluate its need again when MDL-84559 would be finally backported in Moodle core.

lucaboesch:
MDL-84559 ~~is integrated now~~ has been closed as "Won't Fix" since the version to which this is backported (4.5) is now out of general support.

abias:
I think it's a pity that MDL-84559 remained unhandled in Moodle tracker for so long that it was now closed as "Won't fix".

Luca has just proposed internally to just add this change to BU 4.5 to help admins who still run Moodle 4.5 LTS for the time being.

I would not oppose to do that, but my 3 change requests from above are still valid.

Additionally, I think that we will need to add an upgrade step in BU 5.0 and 5.1 which silently checks if the BU setting of this feature here is set and, if yes, will silently activate $CFG->navfilter in Moodle core from 5.0 on. Otherwise, admins who just update from Moodle + BU 4.5 to Moodle + BU 5.x would suffer from a regression that navbar filtering is off again.

Cheers,
Alex

PR #1072, opened 2025-10-20
Tests: Add tests for Boost Union features which might but must not affect other active themes, resolves #761

### LINKED ISSUE #761: Add tests for Boost Union features which might but must not affect other active themes
In multiple places, Boost Union is checking if it is the active theme with the `theme_boost_union_is_active_theme()` function. This is done to avoid that the feature implementation affects other active themes.

It might be worthwhile to have tests which cover these checks and make sure that the feature is really not affecting other active themes.

abias:


abias:
The tests have passed. The failing tests are unrelated to this PR.

PR #1112, opened 2026-01-06
Add new login layouts and other login settings, resolves #1055

### LINKED ISSUE #1055: Tabbed login paragraphs
Currently there are several login paragraphs (providers/possibilities) below each other (we can change the order, jay!) on the login site. But there is still one very noticeable login form (the one for local users) that our users love to hit their credentials in without thinking/reading.

There is a hack around putting every login paragraph for the different providers in a separate tab with the most popular as active one:
ImageImage

This way your login method is still quickly available but users are more guided to the correct one. I'd like to see this hack incoperated in the official boost union plugin.

We checked _local_login_ plugin but were not convinced that this is our way. Opening a pop-up (even in the browser) seems to be intrusive and the for the proposed way changes should be relatively simple.

berthob98:
### Description

This PR introduces two new login layouts **tabs** and **accordion** along with a new login background layout option, allowing users to switch between the default layout and a split-screen layout. Additionally, it adds settings to control various aspects of the login page, including the login container width, the login texts, and the option to disable individual login methods. Resolves #1055 

**Example layouts:**
- **Tabs with default background:**
image

- **Accordion with split-screen background:**
image

### Added Settings:

1. **Login container width** Controls the width of the login container (default: 500px) `logincontainerwidth`

2. **Login layout** Choose between vertical, tabs, or accordion layout `loginlayout`

3. **Login background layout** Choose between default or split screen layout `loginbackgroundlayout`

4. **Primary login** Specify which login method should be active by default in tabs/accordion layouts `primarylogin`

5. **Local login intro text** Customize the intro text for local login `loginlocalintrotext`

6. **Local login tab text** Customize the tab text for local login (tabs layout only) `loginlocallogintabtext`

7. **IDP login** Control if IDP login buttons are shown `loginidploginenable`

8. **IDP login intro text** Customize the intro text for IDP login `loginidpintrotext`

9. **IDP login tab text** Customize the tab text for IDP login (tabs layout only) `loginidplogintabtext`

10. **Self registration** Control if self registration is shown `loginselfregistrationenable`

11. **Self registration intro** Control if intro text is shown for self registration `loginselfregistrationshowintro`

12. **Self registration intro text** Customize the intro text for self registration `loginselfregistrationintrotext`

13. **Self registration tab text** Customize the tab text for self registration (tabs layout only) `loginselfregistrationlogintabtext`

14. **Guest login** Control if guest login is shown `loginguestloginenable`

15. **Guest login intro** Control if intro text is shown for guest login `loginguestshowintro`

16. **Guest login intro text** Customize the intro text for guest login `loginguestintrotext`

17. **Guest login tab text** Customize the tab text for guest login (tabs layout only) `loginguestlogintabtext`

### Technical changes:

The code in this PR prepares the `loginmethods` inside [core_renderer.php](https://github.com/berthob98/moodle-theme_boost_union/blob/ddd213c89d7c876c273aad6a03c3b295d9ba1cc3/classes/output/core_renderer.php) and iterates over these methods inside the [loginform.mustache](https://github.com/berthob98/moodle-theme_boost_union/blob/ddd213c89d7c876c273aad6a03c3b295d9ba1cc3/templates/core/loginform.mustache). This ensures that the login methods are rendered in the preferred order in the HTML DOM. As a result, Bootstrap `nav-tabs` and accordion classes can be easily applied.

Inside [loginform.mustache](https://github.com/berthob98/moodle-theme_boost_union/blob/ddd213c89d7c876c273aad6a03c3b295d9ba1cc3/templates/core/loginform.mustache), the flags `loginaccordion`, `loginbackgroundlayoutdefault`, and `logintabs` are used to conditionally render the appropriate elements for the different layouts.

The split-screen layout is implemented via CSS by applying a new class: `login-background-layout-splitscreen`.

### DIFF
--- a/classes/output/core_renderer.php
+++ b/classes/output/core_renderer.php
@@ -836,7 +836,7 @@ class core_renderer extends \theme_boost\output\core_renderer {
         );
 
         // Compute show* flags for all four login types (theme setting + Moodle core).
-        // visibility is controlled in the template via these show* parameters.
+        // Visibility is controlled in the template via these show* parameters.
 
         // Local login: theme setting only.
         $loginlocalloginsetting = get_config('theme_boost_union', 'loginlocalloginenable');
@@ -851,14 +851,16 @@ class core_renderer extends \theme_boost\output\core_renderer {
             ? $loginidploginenablesetting
             : THEME_BOOST_UNION_SETTING_SELECT_YES;
         $context->showidplogin = ($showidploginenabled == THEME_BOOST_UNION_SETTING_SELECT_YES) &&
-            !empty($context->hasidentityproviders) && !empty($context->identityproviders);
+            !empty($context->hasidentityproviders) &&
+            !empty($context->identityproviders);
 
         // Guest login: theme setting AND Moodle core guest login button enabled.
         $loginguestloginenablesetting = get_config('theme_boost_union', 'loginguestloginenable');
         $showguestloginenabled = ($loginguestloginenablesetting != false) ?
             $loginguestloginenablesetting : THEME_BOOST_UNION_SETTING_SELECT_YES;
         $coreguestloginbutton = !empty(get_config('core', 'guestloginbutton'));
-        $context->showguestlogin = ($showguestloginenabled == THEME_BOOST_UNION_SETTING_SELECT_YES) && $coreguestloginbutton &&
+        $context->showguestlogin = ($showguestloginenabled == THEME_BOOST_UNION_SETTING_SELECT_YES) &&
+            $coreguestloginbutton &&
             !empty($context->canloginasguest);
 
         // Self registration: theme setting AND Moodle core registerauth configured.
@@ -870,8 +872,9 @@ class core_renderer extends \theme_boost\output\core_renderer {
             && $coreregisterauth
             && (!empty($context->cansignup) || !empty($context->hasinstructions));
 
-        // Process intro settings only when the corresponding login type is shown.
+        // Compute intro settings, but only when the corresponding login type is shown.
 
+        // Local login.
         if ($context->showlocallogin) {
             $loginlocalshowintrosetting = get_config('theme_boost_union', 'loginlocalshowintro');
             $showlocalloginintro = ($loginlocalshowintrosetting != false) ?
@@ -880,15 +883,12 @@ class core_renderer extends \theme_boost\output\core_renderer {
                 $context->showlocalloginintro = true;
                 $loginlocalintrotext = get_config('theme_boost_union', 'loginlocalintrotext');
                 if (!empty($loginlocalintrotext)) {
-                    $context->localloginintrotext = format_string(
-                        $loginlocalintrotext,
-                        true,
-                        ['context' => context_system::instance()]
-                    );
+                    $context->localloginintrotext = format_string($loginlocalintrotext);
                 }
             }
         }
 
+        // IDP login.
         if ($context->showidplogin) {
             $loginidpshowintrosetting = get_config('theme_boost_union', 'loginidpshowintro');
             $showidploginintro = ($loginidpshowintrosetting != false) ?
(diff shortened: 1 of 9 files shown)


lib.php
```diff
@@ -336,6 +339,14 @@ function theme_boost_union_get_pre_scss($theme) {
     if (get_config('theme_boost_union', 'blockdrawerwidth')) {
         $scss .= '$drawer-right-width: ' . get_config('theme_boost_union', 'blockdrawerwidth') . ";\n";
     }
+    // Set variables which are influenced by the logincontainerwidth setting.
```
abias: This is the section for SCSS variables which are already present in Boost core and which do not need a default value to be set.

The logincontainerwidth variable should be moved further down to the "custom Boost Union SCSS variable" section.

I will do that in a review commit.

settings.php
```diff
@@ -995,24 +995,100 @@
...
+        $title = get_string('loginlayoutheading', 'theme_boost_union', null, true);
+        $setting = new admin_setting_heading($name, $title, null);
+        $tab->add($setting);
+
+        // Setting: Login layout.
+        $name = 'theme_boost_union/loginlayout';
+        $title = get_string('loginlayoutsetting', 'theme_boost_union', null, true);
+        $description = get_string('loginlayoutsetting_desc', 'theme_boost_union', null, true);
+        $loginlayoutoptions = [
+            'vertical' => get_string('loginlayoutvertical', 'theme_boost_union'),
```
abias: The keys of this array should use constants, just like the other settings in BU.

I will fix that in a review commit.

lang/en/theme_boost_union.php
```diff
@@ -317,9 +317,16 @@
 // ... ... Setting: login form transparency.
 $string['loginformtransparencysetting'] = 'Login form transparency';
 $string['loginformtransparencysetting_desc'] = 'With this setting, you can make the login form slightly transparent to let the background image shine through even more.';
+// ... ... Setting: Login container width.
```
abias: Unfortunately, the new strings are not aligned with the order of the settings in settings.php which makes it harder to keep the overview over the strings in this large language pack. And you introduced some strings which are not used at all.

I will fix that in a review commit.

layout/login.php
```diff
@@ -46,6 +53,7 @@
     'logincontainerclass' =>
             (get_config('theme_boost_union', 'loginformtransparency') == THEME_BOOST_UNION_SETTING_SELECT_YES) ?
                     'login-container-80t' : '',
+    'loginbackgroundlayoutclass' => $loginbackgroundlayoutclass,
```
abias: I think we should stick to the same syntax like logincontainerclass here.

I will change that in a review commit.

templates/core/loginform.mustache
```diff
@@ -99,33 +120,147 @@
...
+
+{{! =========================================================================== }}
+{{! HEADER SECTION }}
+{{! =========================================================================== }}
 <div class="loginform">
     {{#logourl}}
         <div id="loginlogo" class="login-logo">
             <img id="logoimage" src="{{logourl}}" class="img-fluid" alt="{{sitename}}"/>
-            <h1 class="login-heading visually-hidden">{{#str}} loginto, core, {{sitename}} {{/str}}</h1>
+            <h1 class="login-heading text-break w-100 visually-hidden">{{#str}} loginto, core, {{sitename}} {{/str}}</h1>
```
abias: Adding these classes just in this way without reacticting to a particular BU setting violates BU's main design principle not to change any Moodle core behaviour by default.

Can you please double-check for which of the new settings you need this class and only add it if the setting is set?

scss/boost_union/post.scss
```diff
@@ -332,6 +332,75 @@
...
+
+/*---------------------------------------
+ * Login container positioning for tabs
+ --------------------------------------*/
+
+/* Keep login container vertically centered on page load, but allow it to grow downward only.
+   The container starts centered, but when content changes (tabs switch), it grows downward
+   from that position without shifting the top. */
+@include media-breakpoint-up(md) {
+    .login-wrapper:has(.loginform #login-tabs) {
```
abias: I am slightly concerned about the performance costs of this multiple use of :has here. 
See https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has#performance_considerations.

On the one hand, :has is a quite expensive problemsolver and we should not use it multiple times without real need.
On the other hand, :has will be evaluted here only if a .login-wrapper element is found which is only the case on the login page and not throughout the whole moodle.

Having said that: If you see a change for a small refactoring to inject a .hastabs class into login.mustache (which is already overwritten by BU as well), then please do it. If it gets too complicated, leave it as it is.

classes/output/core_renderer.php
```diff
@@ -612,15 +612,364 @@ public function render_login(\core_auth\output\login $form) {
         if ($showlocalloginintro == THEME_BOOST_UNION_SETTING_SELECT_YES) {
             // Add marker to show the local login intro to template context.
             $context->showlocalloginintro = true;
+            // Check if custom intro text is set.
+            $loginlocalintrotext = get_config('theme_boost_union', 'loginlocalintrotext');
+            if (!empty($loginlocalintrotext)) {
+                $context->localloginintrotext = format_string(
+                    $loginlocalintrotext,
+                    true,
+                    ['context' => context_system::instance()]
```
abias: Passing a context to format_string() is not usual.
If this parameter is not provided, format_string will simply pick the context from $PAGE->context. And on the login page, this is already the system context.

I will remove that parameter in a review commit.

classes/output/core_renderer.php
```diff
@@ -612,15 +612,364 @@ public function render_login(\core_auth\output\login $form) {
...
+            }
+
+            // Sort login methods by order setting.
+            usort($loginmethods, function ($a, $b) {
+                return $a->order <=> $b->order;
+            });
+
+            // Mark the first method in the sorted array.
+            if (!empty($loginmethods)) {
+                $loginmethods[0]->isfirst = true;
```
abias: How about initializing this isfirst attribute already with false then the login methods are added to the array to make the data structure cleaner?

classes/output/core_renderer.php
```diff
@@ -612,15 +612,364 @@ public function render_login(\core_auth\output\login $form) {
...
+                $loginlayout == THEME_BOOST_UNION_SETTING_LOGINLAYOUT_TABS ||
+                    $loginlayout == THEME_BOOST_UNION_SETTING_LOGINLAYOUT_ACCORDION
+            ) {
+                foreach ($loginmethods as $method) {
+                    $label = '';
+                    switch ($method->name) {
+                        case 'local':
+                            $label = get_config('theme_boost_union', 'loginlocallogintablabel');
+                            if ($label === false || empty($label)) {
+                                $label = 'Local login'; // Default.
```
abias: You must not hardcode strings here. Take the string from the language pack instead.

If you do that, you can replace that lenghty switch with a loop which just iterates over all login types.

scss/boost_union/post.scss
```diff
@@ -332,6 +332,75 @@
...
+        min-width: $logincontainer-width;
+        max-width: 100%;
+    }
+}
+
+/*---------------------------------------
+ * Login container positioning for tabs
+ --------------------------------------*/
+
+/* Keep login container vertically centered on page load, but allow it to grow downward only.
```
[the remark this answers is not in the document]
berthob98: I adjusted this slightly by lowering the distance and reducing the minimum distance.

scss/boost_union/post.scss
berthob98: I left this as it is.

PR #1242, opened 2026-04-29
Tests: Drop 'And Behat debugging is (en|dis)abled' steps

### LINKED ISSUE #1241: Tests: Drop 'And Behat debugging is disabled' steps
After https://moodle.atlassian.net/browse/MDL-76975 has been integrated to Moodle Core 5.0 and 5.1, the `Error: too much data passed as arguments to js_call_amd` errors which appeared in the site administration on several BU settings pages should be gone.

Against this background, we could drop the 'And Behat debugging is disabled' and 'And Behat debugging is enabled' steps from all of our Behat scenarios, but only from 5.0 on.

SAY-5:
## Summary

Closes #1241.

Once [MDL-76975](https://moodle.atlassian.net/browse/MDL-76975)
landed in Moodle Core 5.0+, the
\`Too much data passed as arguments to js_call_amd\` debug message
that the workaround was guarding against is gone, and the explicit
\`And Behat debugging is (disabled|enabled)\` steps can be removed
from our scenarios.

## Changes

- Drop \`Behat debugging is (disabled|enabled)\` lines from every
  \`.feature\` file under \`tests/behat/\` (15 files, 110 lines —
  both active steps and the few \`#\`-commented variants).
- Remove the now-unused \`disable_behat_debugging\` /
  \`enable_behat_debugging\` helper methods and their
  \`@Given /^Behat debugging is (disabled|enabled)$/\` annotations
  from \`tests/behat/behat_theme_boost_union_base_general.php\`.

The helper deletion and the scenario edits land in the same commit.
With the helpers gone, Behat would otherwise refuse to recognise any
leftover step text — and any forgotten step text would show up as a
test failure.

## Branch scope

This branch (\`main\`, \`v5.1-r11\`) targets Moodle 5.1, which the
issue lists as in-scope ("affecting 5.1 only" label). Per the issue
the same change applies to the 5.0 and 5.2 stable branches; happy
to backport if that's the convention here, or you can cherry-pick.

## Test plan

- [x] \`php -l tests/behat/behat_theme_boost_union_base_general.php\`
      reports no syntax errors.
- [x] \`grep -r 'Behat debugging is' tests/behat\` is empty.
- [x] \`grep -r 'disable_behat_debugging\\|enable_behat_debugging' tests/behat\`
      is empty.
- [ ] Behat run against a local Moodle 5.0+ — I don't have a Moodle
      install wired up here so I'm relying on CI for the runtime
      check. The change is purely deletion (no new steps, no
      reorder), so the only failure modes are (a) a forgotten step
      text in a scenario, which the grep above rules out, or (b) a
      scenario that genuinely depended on the debug toggle to not
      hit MDL-76975 — but that's exactly what the maintainer's
      issue body says is now safe to drop.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2026-05-01 - Tests: Drop 'And Behat debugging is (en|dis)abled' steps, resolves #1242
 * 2026-04-30 - Cleanup: Drop overwritten templates/theme_boost/drawer.mustache, resolves #1246
 
 ### v5.1-r11
--- a/tests/behat/behat_theme_boost_union_base_general.php
+++ b/tests/behat/behat_theme_boost_union_base_general.php
@@ -475,6 +475,28 @@ class behat_theme_boost_union_base_general extends behat_base {
     }
 
     /**
+     * Disables debugging in Behat.
+     *
+     * We might need to deactivate debugging for a while when Behat steps fail due debugging message which can't be avoided.
+     *
+     * @Given /^Behat debugging is disabled$/
+     */
+    public function disable_behat_debugging() {
+        set_config('debug', 0);
+        set_config('debugdisplay', 0);
+    }
+
+    /**
+     * Re-enables debugging in Behat.
+     *
+     * @Given /^Behat debugging is enabled$/
+     */
+    public function enable_behat_debugging() {
+        set_config('debug', 32767);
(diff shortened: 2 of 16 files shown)


abias:
Thanks @SAY-5 for your contribution!

Your solution is spot-on. I just rebased your branch and re-added the Behat step definitions (as we do not know if we might need something like that later again).

Cheers,
Alex

Issue #1243, opened 2026-04-29
Upgrade: Moodle 5.2


abias:
Same procedure as every 6 months:
Boost Union should be upgraded to 5.2

Richardvi:
Thanks @abias for the great work. You guys rock. BTW, our University would probably love to join the Moodle Hochschule, unfortunately there seems to be a big language barrier... in the Netherlands a lot of Universities are using English as standard. German is not my best foreign language... sorry :)

abias:
Hi @Richardvi ,

> Thanks @abias for the great work. You guys rock. 

You're welcome :)

> BTW, our University would probably love to join the Moodle Hochschule, unfortunately there seems to be a big language barrier... in the Netherlands a lot of Universities are using English as standard. German is not my best foreign language... sorry :)

Well, that's great to hear!
We are open to foreign members of course and would appreciate your membership. But during the foundation of Moodle an Hochschulen e.V., the main language was set to german on purpose and I think it won't be changed to english.

I have raise that case within the board and think we might discuss it in one of the next meetings.

Cheers,
Alex

abias:
Regarding the matter of the 5.2 upgrade:

The working branch is there on https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/tree/upgrade-502 and we are now working through the upgrade changes.

carlmaegermangroeipunt:
Hoi Abias, any idea when Boost Union 5.2 will be released? Our school wants to upgrade to Moodle 5.2 on 01-07-2026. We hope that your team can do wonders ;-)

abias:
Hi @carlmaegermangroeipunt ,

we are currently working through the upgrade work and are making good progress. Just the new login page is a hard nut as we have to integrate all of our settings and improvements.

You can help us by testing the https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/tree/upgrade-502 branch and report any bugs and glitches which you encounter (except on the login page ;) ) into new issues.

Cheers,
Alex

abias:
To everyone who is watching this issue:

You should be ready to play with the  https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/tree/upgrade-502  branch on your testing instances.

We have finished most of the blocking codework, including support for the new Moodle login page layout. Our manual tests will start soon after which we will publish the first 5.2 release.

If you encounter any additional problems or glitches with BU on 5.2, please raise a new issue here on Github.

[...]

PR #1244, opened 2026-04-30
Improvement: Make OAuth2 login button color configurable, resolves #1225

### LINKED ISSUE #1225: Improvement: Make OAuth2 login button color configurable
Currently, there is a deviation in the color of the OAuth2 login buttons between Boost Core and Boost Union:

Boost Core:

Image

Boost Union:

Image

This deviation was introduced in #473 on purpose.
But it was not added to https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/blob/main/README.md#exceptions-to-our-main-design-principle unfortunately.

We should at least mention the deviation in the README.
But even better, there should be an admin setting to allow the admin to decide himself if he want to have the color change or not.

rohithvadla99:
This PR implements the fix proposed in issue #1225.

**Changes made:**

- `settings.php`: Added new admin setting `loginidpbuttoncolorenable` under the
  Login → IDP section, using the same `admin_setting_configselect` / `hide_if` 
  pattern as adjacent IDP settings. Default is Yes to preserve existing behavior
  for all current Boost Union installations.

- `lib.php`: Added conditional SCSS injection in `theme_boost_union_get_extra_scss()`.
  When the setting is enabled, the `.login-identityproviders .btn` primary color
  override is compiled in. When disabled, no override is injected and Boost core's
  default unstyled appearance is restored.

- `lang/en/theme_boost_union.php`: Added the two required language strings for the
  new setting title and description.

- `README.md`: Documented the OAuth2 button color deviation under the
  "Exceptions to our main design principle" section, as requested in the issue.

**Testing done:**
- Setting to Yes → primary color applied to IDP buttons on login page ✓
- Setting to No → buttons revert to Boost core default (white/unstyled) ✓
- Cache purge after each toggle forces SCSS recompile correctly ✓
- `php -l lib.php` and `php -l settings.php` pass with no syntax errors ✓


lucaboesch:
This relies on #1243 which introduces the primarylight lang string and such.
That is why the tests are failing right now.

lucaboesch (changes requested):
Thanks for your pull request!
As mentioned in a comment, this relies on some strings existing and should be merged only when those are in, they come with #1243.
But the thing you could do is the following: make sure the Pull request consists only of one commit, "Improvement: Add admin setting to toggle OAuth2 IDP login button color, resolves 1225", so it builds up directly on the main branch.
Thanks for your hard work!

rohithvadla99:
That makes sense! I was wondering why the tests were failing. Also, thanks for the input. Will do that!

abias:
Hi @rohithvadla99 ,

I am really sorry, but I had to supersede this PR with a PR of my own in #1259.

This was done on the one hand because your patch was apparently not rebased correctly unfortunately and undid some of the latest changes in Boost Union (see https://github.com/moodle-an-hochschulen/moodle-theme_boost_union/pull/1244/files). Thus, it would have needed some more work.

But it was mainly done because I am working on the login page buttons myself now in preparation for the Moodle 5.2 upgrade and I needed to introduce some more settings beyond your patch.

Anyway, I added you to the Credits section in README.
I would appreciate if you would pick another "good first issue" and just contribute something else :)

Cheers,
Alex

PR #1265, opened 2026-05-15
Fix #1063: rename remaining 'courselistinghow*' settings to 'courselistingshow*'

### LINKED ISSUE #1063: Glitch: Typo in courselistinghow* strings and settings
I just became aware that there was a glitch in our QA process.
If you search the Boost Union codebase for `courselistinghow`, you will find several matches.
All of these should correctly say `courselistingshow` instead.

While this may seem like a simple search-and-replace task, it is not. 
To fix this typo, we would also have to create update scripts for the changed settings names and have to accept that translations will have to be re-done in AMOS.

Oksigenia:
This PR resolves #1063.

Background:
The `courselistinghowfields` setting was renamed to `courselistingshowfields` in the 2025041413 upgrade step to correct a typo. The other seven settings of the same family were left untouched at that time.

This PR completes the cleanup by renaming the remaining settings:

- `courselistinghowimage` -> `courselistingshowimage`
- `courselistinghowshortname` -> `courselistingshowshortname`
- `courselistinghowcategory` -> `courselistingshowcategory`
- `courselistinghowprogress` -> `courselistingshowprogress`
- `courselistinghowenrolicons` -> `courselistingshowenrolicons`
- `courselistinghowgoto` -> `courselistingshowgoto`
- `courselistinghowpopup` -> `courselistingshowpopup`

The migration follows the same pattern used in the existing 2025041413 upgrade step, but consolidates all seven settings into a single upgrade block to keep the file readable.

Files changed:

- `lang/en/theme_boost_union.php` - renamed string keys plus a new `upgradenotice_2026051500` string
- `settings.php` - renamed `admin_setting_configselect` names and the two `hide_if` dependencies that point to `courselistinghowimage` and `courselistinghowprogress`
- `classes/output/core/course_renderer.php` - updated `get_config()` calls
- `db/upgrade.php` - new 2026051500 upgrade step
- `version.php` - version bump
- `CHANGES.md` - unreleased entry

About AMOS / translations:
As noted by @abias in the issue thread, AMOS does not parse Git logs for plugins, so existing translations of the renamed strings will need to be re-done. This PR does not handle translations; the string keys are simply renamed in `lang/en`.

Testing:
Verified on a live Moodle installation by setting all seven affected settings to non-default values, applying the patch, running the upgrade and confirming that:

- All seven old config keys are removed from `mdl_config_plugins`
- All seven new keys are created with the original values
- The upgrade notice is shown to the admin
- The course listing renders correctly with the renamed settings

The plugin's `$plugin->supported` declaration is left untouched.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changes
 
 ### Unreleased
 
+* 2026-05-15 - Chore: Rename the settings 'courselistinghow*' to 'courselistingshow*' to fix a typo, resolves #1063
 * 2026-05-12 - Improvement: Allow the admin to further control the logo, heading, and tagline presentation on the login page, resolves #1262
 * 2026-05-12 - Improvement: Allow the admin to set the login page logo max-width and max-height as well as a margin-bottom and the horizontal alignment for the logo, resolves #1260
 * 2026-05-12 - Improvement: Add admin settings to set the button colors on the login page, resolves #1225.
--- a/classes/output/core/course_renderer.php
+++ b/classes/output/core/course_renderer.php
@@ -49,12 +49,12 @@ class course_renderer extends \core_course_renderer {
         static $detailsmodalchecked = null;
         if ($detailsmodalchecked == null) {
             $courselistingpresentation = get_config('theme_boost_union', 'courselistingpresentation');
-            $courselistinghowpopup = get_config('theme_boost_union', 'courselistinghowpopup');
+            $courselistingshowpopup = get_config('theme_boost_union', 'courselistingshowpopup');
             if (
                 isset($courselistingpresentation) &&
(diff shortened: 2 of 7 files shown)


abias:
Hi @Oksigenia ,

many thanks for submitting this PR!
Your patch looks well structured and basically ready.

I just would like to postpone this change to BU 5.3 LTS as it is just an internal change and I do not want to burden the AMOS translators with the re-translation effort currently as we have introduced so much new strings in the last weeks already.

So I will have another look and rebase the patch in roughly 5 months.

Thanks for your work and your understanding.

Cheers,
Alex

Oksigenia:
Hi Alex,

Sounds good, thanks for the context on the AMOS planning. The cleanup can sit until 5.3 — happy to wait for your rebase, or to refresh the branch myself closer to that release if it helps. Just ping me.

Cheers,
Oksigenia

PR #1336, opened 2026-06-21
Bugfix: Filter course names in starred courses popover, resolves #1332

### LINKED ISSUE #1332: String filters (e.g. the multilang filter) not applied to the course names in starred courses popover

lucaboesch:


templates/popover-favourites.mustache
```diff
@@ -54,7 +54,7 @@
 
     {{$content}}
         {{#favourites}}
-            <a class="dropdown-item {{^visible}}dimmed{{/visible}}" href="{{{url}}}" title="{{{fullname}}}">{{{fullname}}}</a>
+            <a class="dropdown-item {{^visible}}dimmed{{/visible}}" href="{{{url}}}" title="{{{fullname}}}">{{fullname}}</a>
```
abias, line 57: I stumble upon the fact that you are using three brackets in the title attribute and two in the link text.
This means that the fullname still lands in its unescaped form in the title tag.
This isn't a problem since you used `clean_string()` before. But it's still inconsistent.

Shouldn't we also use two brackets in the title attribute?
And if we do, do we then still need `clean_string()`?

tests/behat/theme_boost_union_feelsettings_navigation.feature
```diff
@@ -10,8 +10,8 @@ Feature: Configuring the theme_boost_union plugin for the "Navigation" tab on th
       | student1 |
       | teacher1 |
     And the following "courses" exist:
-      | fullname | shortname |
-      | Course 1 | C1        |
+      | fullname                                                                                                 | shortname |
+      | <span lang="en" class="multilang">Course</span><span lang="de" class="multilang">Kurs</span> 1 & < ' " > | C1        |
```
abias: My gut feeling is that we should rather create and use a dedicated course with multilang name in the `Setting: Show starred courses popover in the navbar` scenario instead of using this complicated course name here for the whole feature file.

And if we do that, we do not need to enable the multilang filter in each scenario anymore.

abias (changes requested):
Hi @lucaboesch ,

many thanks for fixing this bug!
You patch is understandable, but I have two refinement comments and I would appreciate if you could handle them as, as it stands now, your patch seems to be more complicated than necessary ;)

templates/popover-favourites.mustache
lucaboesch: It's the same exact behaviour of Moodle core (the fixes https://moodle.atlassian.net/browse/MDL-77364 and https://moodle.atlassian.net/browse/MDL-86830), reasonings can be taken from https://moodle.atlassian.net/browse/MDL-87066.
I'm following the examples, here.
I understand it's unfortunately necessary.
https://moodle.atlassian.net/browse/MDL-79755 (currently in review) is doing this the same way, too.

tests/behat/theme_boost_union_feelsettings_navigation.feature
lucaboesch: Hhmh I judged you the other way around!
I believed you wanting to activate the multilang filter the less possible, to save computing power. That means some lines more, right. But it probably is somewhat (unnoticeably) faster.

lucaboesch:
I reworked the tests/behat/theme_boost_union_feelsettings_navigation.feature Behat test.
Thanks for pointing that out!

Best,
Luca

PR #1373, opened 2026-07-17
Improvement: For course header images, allow the teacher to adopt the existing course image as course header image, resolves #1371

### LINKED ISSUE #1371: Course header: Allow the user to use the existing course image for the course header image too
**Is your feature request / improvement related to a problem? Please describe.**
* It is possible, that users have already uploaded a course image to a course but do no have a course header image yet. 
* There is a configuration option that sets the use of image fallbacks in a way, that the course image is **not** set to automatically to be also used as the course header image, which is totally reasonable, since not all course images will fit the necessary format to be used as such.
* but there may be cases, in which a teacher decides, he still wants to use the same image (maybe another image detail) also as the course header image. In this case it would be very helpful to have a button that enables to user to use the course image also as course header image without having to download and upload the same picture again. 

Image

**Describe the solution you'd like**
This could be solved via a button in the upload interface (use course image)

**Describe alternatives you've considered**

**Additional context**

abias:
This is how the course images form on the course settings page looks like as long as no course image has been uploaded:

Bildschirmfoto 2026-07-17 um 15 50 55

As soon as you upload a course image (or if a course image is already uploaded when the form is loaded), the new button appears:

Bildschirmfoto 2026-07-17 um 15 51 30

Clicking the button brings up a confirmation dialogue:

Bildschirmfoto 2026-07-17 um 15 51 11

And after confirming the adoption, the image is copied over to the course header image area and the button disappears again:

Bildschirmfoto 2026-07-17 um 15 51 19


wiebkemueller-hsh:
Hello Alex, I sucessfully tested the feature and it works exactly as we agreed with (maybe?) one edge case. I wonder if this is realistic. In theory, if during the editing process the user would delete the course image, the button remains but has no effect, obviously. The button disappears soon after the course settings are saved. So I think this is neglectable, but since I noticed it I thought I might as well write it down. 
Thank you! Wiebke

abias:
Hi @wiebkemueller-hsh ,

thank you for your test and your feedback.
I think your edge case is valid and I will push an updated commit which handles the edge case.

PR #1379, opened 2026-07-19
 Bugfix: Prevent that rendering the primary-drawer-mobile-child mustache template exceeds the allowed memory size, resolves #874

### LINKED ISSUE #874: Mustache Cache Error after fresh install of v4.5-r10
**Describe the bug**
I got the following error after installing v4.5-r10 on Moodle 4.5.2+ (Build: 20250308):
"Fatal error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 315392 bytes) in /var/www/moodledata/localcache/mustache/1742624452/boost_union/__Mustache_f88dc471bd71c619dff5222848f518e8.php on line 59"

**To Reproduce**
1. Go to '(https://moodle.org/plugins/theme_boost_union/versions)'
2. Click on 'Get the latest version for moodle 4.5'
3. Install the plugin over sftp
4. Go to the admin settings and see the error code

**Expected behavior**
Bevor I used the 4.1 Moodle Version and upgraded Moodle to 4.5.2. After that I started a fresh install of Moodle Boost Union. PHP 8.3 is working fine and I also raised the php cli and php apache2 memory_limit to 2048 M to check. But nothing changed. I got the mustache error.

**Screenshots**

 - Moodle 4.5.2+ (Build: 20250308):
 - Boost Union Version
 - Linux Ubuntu 
 - Browser [Firefox, Brave, Chrome]
 - ...

**Additional context**

abias:


abias:
Hi @prasanna-lmsace ,

this is a mitigation for #874 which Claude built for me. 
As #874 was not really reproducible, it was not really easy to identify the original problem and to build a fix.
I would be really grateful if you could have a look if we can integrate this patch or if you rather think that this is nonsense.

Thanks,
Alex

